PR #9, opened 2015-12-03
Fixed the bug with the subtraction.

mrakgr:
Fixed the bug with the subtraction in issue 8.


gbaydin:
The fix is correct, thank you for submitting this!

There are also two associated bugs in `Sub_D_DV` and `Sub_D_DM`, I will commit the fix for them subsequently.

PR #10, opened 2015-12-13
Master

mrakgr:
Except for gesv the solver functions were pretty much all broken. I
fixed it and added some notes on their design. There is an issue that
gesv and sysv are not mutable while getrf and getri are. I've also made sysv
non mutable because I noticed that it was not being backed up.

Except for gesv all of them needed transposition steps at the beginning
and at the end.

Github ate the description I left with the pull request, so let me try to reproduce it...

Besides the above issues, I left some comments in the backend file for you to peruse. I've solved the fact that the functions were not working, but I've left the design issue for you to think about. I'll continue writting the Cuda backend from here. Most of the necessary pieces I've already made in Alea, I just need to translate them into C++. The only novel thing I will have to do is make elementary functions for broadcasting with row major order.

After all this is done, it might be worth transporting the library to column major order as both OpenBLAS and Cuda libraries use it natively.

Just today, I've noticed that OpenBLAS has solver functions that take order as an argument, but when I tried it I stumbled upon some library bug that causes the runtime to crash. It worked fine with column major order so I am sure I got the parameters right, but not with row major.

Well, have fun.


gbaydin:
Dear @mrakgr, thank you very much for your interest in the library and working on fixes and enhancements. :)

Unfortunately, I cannot merge this pull request as it is, because it involves modifications in an essential part of the library (`Backend.OpenBLAS.fs`) and also adds several `fsx` script files that we do not want to be placed within the library, due to our way of organizing the project. (We prefer to use `fsx` files for library documentation, under the folder `docs/input`).

In the description above, you mention that there are possibly some bugs in the OpenBLAS backend. If you encounter a bug, can you please raise separate issues where we can focus just on the bug, with a minimal working example? If you want to contribute fixes for any bug, can you please submit minimal pull requests with commits only fixing that particular bug?

About your interest in contributing to a CUDA backend, I think it would be best if we look at it using a pull request involving a clearly separated `Backend.Cuda.fs` file (like you already have here in this pull request). If your code seems to be working, passes the unit tests, and has performance not worse than the OpenBLAS backend, we can accept it into a separate experimental CUDA branch. (A correctly implemented CUDA backend should be some orders of magnitude faster than our CPU backend.)

I am very happy to read all your comments here in the code and elsewhere. I will certainly follow your progress and have a look at how your `Backend.Cuda.fs` is working. Thank you!

mrakgr:
Hmmm, I am not too familiar with Github, so I guess doing this all at once was too much. I had assumed that it would be fairly obvious at least that the key transposition steps were missing in three of the functions if one looked at the changes.

I'll go step by step then and raise all the issues separately tomorrow as you'd prefer.

About that script file, I know I should not have put it in, but I saw no options for partial pulls, so I had hope that you might opt to just not to add it in. I'll do some more research on how to use Github effectively.

Issue #20, opened 2016-07-19
duplicate implementation of AD.Float32 and AD.Float64


smoothdeveloper:
I was looking at implementation files `AD.Float32.fs` and `AD.Float64.fs` and was thinking I could contribute a refactoring so the code is 100% identical but a dozen of lines at the top of each file.

It would basically consist of defining type alias and few literals. I think it would make extending and maintaining the library easier than it is now.

Does that sound like a good idea?

smoothdeveloper:
Note that it would compile to exactly the same thing.

dsyme:
Making the code identical sounds good

smoothdeveloper:
@dsyme thanks I'll contribute a PR to that aim then.

smoothdeveloper:
@dsyme / @gbaydin I've put a first commit in PR #20 it changes the AD.Float32 module and wonder if you could look at the diff and confirm all the naming is OK to proceed with AD.Float64.

Also, could we get #19 merged, it adds running the tests in our AppVeyor integration which are currently not running in master. master also doesn't build the tests for me which I think I've fixed in my paket PR.

smoothdeveloper:
I've completed the same in AD.Float64, files look the same now besides the few aliases at the top.

dsyme:
@smoothdeveloper Please feel free to contribute further alignments

PR #21, opened 2016-07-20
code alignment for AD.Float32 & AD.Float64

smoothdeveloper:
Pass on AD.Float32 for early review.
- I'm wondering if we need a numberNeg1 (I think no but couldn't verify for sure that constant was introduced)
- I haven't found a way to make added `inline` members internal

Looking forward for feedback on naming / formatting / adjustments before applying same changes to AD.Float64 which should make the files pretty much identical.
#20


smoothdeveloper:
@gbaydin this PR is still mergeable without conflict, I think it will help further maintenance of the library.

PR #28, opened 2017-07-19
cleanup and add build.sh

dsyme:
Was scanning through the library and did some minor cleanup to make the code more readable (whitespace etc.)

Also this fixes a build error in overload resolution incurred with F# 4.1 (VS2017)


dsyme:
I also pushed some changes to 
* get tests to pass on Linux
* take the OpenBlas binaries from the "lib" directory

dsyme:
OK, we now have working CI on Linux.  It took me a while to track down how to install a working libopenblas-dev package on Ubuntu 14.04, I found one here:
```
sudo add-apt-repository ppa:ginggs/julia-backports -y
sudo apt-get update -y
sudo apt-get install libopenblas-dev -y
```

PR #35, opened 2017-07-20
Keep reverse mode AD tape in lookaside datastructure

dsyme:
Hi @gbaydin @barak 

I would like to move to a model where the core D/DV/DM values are immutable, and the reverse-mode AD tape is kept in a lookaside "adjoints" data structure (a dictionary) - and likewise the fanout counts are kept as a local mutable dictionary data structure only active during the reverse AD sweep

Instead of the mutating

```
L |> reverseProp (D 1.) // Propagate adjoint value 1 backward
```

you now have

```fsharp
let adjoints = e |> computeAdjoints // Propagate adjoint value 1 backward
```
This means that the usage model for reverse-mode AD becomes more explicitly value-based and functional, e.g.
```fsharp
open DiffSharp.AD.Float64

// Get a fresh global tag for this run of reverse AD
let i = DiffSharp.Util.GlobalTagger.Next

// Initialize input values for reverse AD
let a = D 0.5 |> makeReverse i
let b = D 1.2 |> makeReverse i

// Perform a series of operations involving the D type
let e = (sin a) * (a + b)

// Propagate the adjoint value of 1 backward from e (or de/de = 1)
// i.e., calculate partial derivatives of e with respect to other variables
let adjoints = e |> computeAdjoints

// Read the adjoint values of the inputs
// You can calculate all partial derivatives in just one reverse sweep!
let deda = adjoints.[a]
let dedb = adjoints.[b]
```
without mutating the actual graph values

My understanding is that this has no significant performance impact since the time is entirely dominated by matrix/vector computations, and in any case the dictionaries used for tape lookaside are very fast.

### DIFF
--- a/Roadmap.txt
+++ b/Roadmap.txt
@@ -1,13 +1,16 @@
-0.7.*
+1.1.*
+- Separate reverse tape from node descriptions
+
+1.2.*
 - Fully generalize differentiation API type signatures
 - GEMV/GEMM ops with vector addition
 - Iterate to fixed point
 - In-place operations
 - CUDA/CuBLAS backend
 
-0.8.*
+1.3.*
 - Tensor/multidimensional array, a la numpy.ndarray
 - Convolution ops
 
-0.9.*
+1.4.*
 - AD via syntax tree transformation
--- a/docs/BuildDocs.fsx
+++ b/docs/BuildDocs.fsx
@@ -41,12 +41,13 @@
 //
 
 #r "../packages/FSharp.Compiler.Service/lib/net45/FSharp.Compiler.Service.dll"
+#r "../packages/FSharp.Formatting/lib/net40/FSharp.Markdown.dll"
 #r "../packages/FSharpVSPowerTools.Core/lib/net45/FSharpVSPowerTools.Core.dll"
+#r "../packages/FSharp.Formatting/lib/net40/RazorEngine.dll"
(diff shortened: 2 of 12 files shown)


dsyme:
The key change to keep the tape in a lookaside table is here: https://github.com/DiffSharp/DiffSharp/pull/35/files#diff-ae1d8d0bc2d022e39b062a17e7229157R3237

barak:
This seems like a good idea. It should in principle make nesting easier, and lends itself to more aggressive optimizations. Ultimately of course the tape should be made out of lambdas, but that would require very different machinery.

The reason for the previous strategy was that it's what R6RS-AD uses, which was driven by ease of idiomatic implementation in Scheme.

dsyme:
> This seems like a good idea. It should in principle make nesting easier

Yes, I was wondering if the implementation of nesting should use different adjoint tapes, but I kept the existing behaviour and used just one for the entire backprop

> and lends itself to more aggressive optimizations. 

Yes, it feels like it opens up new opportunities for parallelism and could make some kinds of distribution simpler

dsyme:
This caused a small perf regression according to dsbench.exe. dsbench.exe mainly  tests "small" AD with lots of D values.  If you increase the vector size the difference gets less and less.

I pushed an update to use faster integer indexing into the dictionary rather than keying on object identity

There is also a truckload of low hanging fruit for perf, see table below for profiling of dsbench.exe 

image

Issue #18, opened 2016-06-20
DM and DV binary operations are not defined in Interop module.


zgrkpnr:
For instance DM \* DM is definet, DV \* DV is also defined. But Matrix and Vector operations such as DM \* DV is not defined.

smoothdeveloper:
@zgrkpnr could you provide a sample F# code which would do what you want from user perspective?

It is nice if the script has checks on the results of those operations so it can be converted to a test and the library implemented according to that spec.

dsyme:
@zgrkpnr I pushed a fix for this.

BTW would you be interested in helping with admin, testing and future development on this project ?  It is an enjoyable codebase to work on!

Issue #23, opened 2016-10-09
Missing some operations on vectors/matrices C#


Zerseu:
Hello,

I am unable to find the following operations on vectors and matrices when coding in C#:

let v = v1 &\* v2 // Vector outer (dyadic, tensor) product
let v = v1 .\* v2 // Element-wise (Hadamard) product
let v = v1 ./ v2 // Element-wise division
let v = v1 *\* v2 // Element-wise exponentiation

And:

let m  = m1 .\* m2 // Element-wise (Hadamard) product
let m  = m1 ./ m2 // Element-wise division
let m  = m1 *\* m2 // Element-wise exponentiation

Thank you.

dsyme:
@Zerseu This is because those operators aren't available in the C# language. https://github.com/DiffSharp/DiffSharp/issues/37 pushes a fix where you can do

```fsharp
DV.PointwiseMultiply(a,b)
DV.PointwiseDivision(a,b)
DV.Pow(a,b)
DV.TensorMultiply(a,b)

DM.PointwiseMultiply(a,b)
DM.PointwiseDivision(a,b)
DM.Pow(a,b)
```

Please send a PR if you would like to change the names if you can identify more standard names in other math libraries

PR #37, opened 2017-07-20
additional interop binary ops

dsyme:
addresses #18


dsyme:
Also fixes this: https://github.com/DiffSharp/DiffSharp/issues/23

PR #50, opened 2019-03-14
Enable building for macOS

### LINKED ISSUE #22: OpenBLAS x86 also works with DiffSharp
I have tried using native providers' x86 builds and used "Any CPU" as build configuration. It builds and runs.

I have testes vector and matrix multiplication, transpose etc. 
gradient and hessian also works. 

I obtained x86 builds of native providers from MathNet distribution. They have both x64 and x86 providers.

### LINKED ISSUE #24: Any chance to have it working on Mac Os X? (re-enable for 32-bit?)
Hi,

I would love to use it on Mac OsX. Any idea what I should do?

currently I get

```
System.DllNotFoundException: libopenblas                                                                                          
  at (wrapper managed-to-native) DiffSharp.Backend.OpenBLAS+BLASExtensions:cblas_somatcopy
```

erydo:
Based on #49 (_Fix incorrectly-placed UTF-16 BOM_).

This branch builds with no test errors on macOS (Mojave 10.14.3, Darwin Kernel Version 18.2.0 x86_64).

Sets builds for 'Any CPU' which allows the solution to build on macOS. Fixes #22.
Adds OpenBLAS@0.3.5 for macOS. Fixes #24.

See more detailed comments in the commits themselves.

Note that the OpenBLAS lib is vendored directly and, like the existing ones, is pretty large. GitHub gave this warning when pushing:
```
remote: warning: GH001: Large files detected. You may want to try Git Large File Storage - https://git-lfs.github.com.
remote: warning: See http://git.io/iEPt8g for more information.
remote: warning: File lib/OpenBLAS-v0.3.5-macOS-x86_64/libopenblas.dylib is 57.31 MB; this is larger than GitHub's recommended maximum file size of 50.00 MB
```

Perhaps this repo should use git LFS (it's pretty easy to set up, but there are free and paid tiers to consider) or find another way of vendoring those binaries. This doesn't seem critical.


dsyme:
Super, thank you!!

Issue #52, opened 2019-05-09
New release?


cgravill:
There hasn't been a release of https://www.nuget.org/packages/DiffSharp/ since 2015 and lots of useful changes have been made since. Are there any plans for a release? Blockers you'd like help with?

I was able to build from source so I can use DiffSharp with .NET Core but I'd really like to share the work with others who are unlikely to build manually.

cgravill:
I've pushed up a pre-release version NuGet for testing: https://www.nuget.org/packages/DiffSharp/0.8.0-beta it'd be great to get any feedback if it works for others.

There have been some API changes since the last release. The documentation has been updated. Once it's fully released we should republish the documentation as well. By way of example:

```fsharp
//DiffSharp >= 0.8.0
let adjoints = computeAdjoints L

//DiffSharp <= 0.7.7
L |> reverseProp (D 1.)
```

and

```fsharp
//DiffSharp >= 0.8.0
let mats' = mats |> Array.map (fun x -> primal (x - eta * adjoints.[x]))

//DiffSharp <= 0.7.7
let mats' = mats |> Array.map (fun x -> primal (x - eta * x.A))
```

cgravill:
There have been a few more minor releases up to: https://www.nuget.org/packages/DiffSharp/0.8.4-beta and we've now been able to train some interesting models.

dsyme:
I am closing this as the focus is now on DiffSharp 1.0 improvements. There is a feed of packages from the CI server if you want to play around with those, or you can build your own from `dev`

@cgravill please do continue to update to 0.8.4 etc.

cgravill:
Sounds good, we have a working stable version that we don't have any immediate need to update.

Very much interested in switching to the `dev` version when we get a chance - the `#r "nuget:` under F# 5.0 is brilliant to see:

https://github.com/DiffSharp/DiffSharp#using-ci-build-packages

PR #53, opened 2019-05-29
WIP: .NET Standard NuGet release

cgravill:
#52 

I've added the metadata from the previous NuGet 0.7.7, with some updates. I put the packing information for the native libraries.

@erydo I saw you've been helping out with the macOS builds. Currently the targets file https://github.com/DiffSharp/DiffSharp/blob/3bf7a97f6d7dbab7c85c9131467df480465e54a8/src/DiffSharp/DiffSharp.targets is from the previous NuGet. Given it auto-references specifically the dlls that'll be annoying on macOS. Does it make sense to conditionally get the dylib you added?

@gbaydin & @dsyme any adjustments that we should make before releasing?

### DIFF
--- a/build.cmd
+++ b/build.cmd
@@ -3,3 +3,4 @@ cls
 dotnet --version
 dotnet build DiffSharp.sln -c release -v:n
 dotnet test tests/DiffSharp.Tests  -c release -v:n
+dotnet pack DiffSharp.sln -c release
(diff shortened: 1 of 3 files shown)


cgravill:
Math.NET has a similar strategy:

https://github.com/mathnet/mathnet-numerics/blob/d6d81576ccb5423d1eb256be16063f50acf59410/build/NativeProvider.targets

They build separate NuGets for each target (so contains exclusively either .dll or .so). Anyone have a strong feeling if multiple NuGets or a more intelligent targets file is the best way to go?

PR #54, opened 2019-06-11
Native dependencies, better behaviour on non-Windows

cgravill:
Only copy the Windows native dlls when on Windows
Adjust the Homebrew command for where openblas is now

### DIFF
--- a/build.sh
+++ b/build.sh
@@ -3,7 +3,7 @@
 case "$(uname -s)" in
 
    Darwin)
-     brew install homebrew/science/openblas
+     brew install openblas
      ;;
 
    CYGWIN*|MINGW32*|MSYS*|Linux)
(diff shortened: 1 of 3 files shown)


cgravill:
Note, for macOS, I've found that the included dylib isn't enough to work on macOS, I get:

```
Unhandled Exception: System.TypeInitializationException: The type initializer for '<StartupCode$difftestcore>.$Program' threw an exception. ---> System.DllNotFoundException: Unable to load shared library 'libopenblas' or one of its dependencies. In order to help diagnose loading problems, consider setting the DYLD_PRINT_LIBRARIES environment variable: dlopen(liblibopenblas, 1): image not found
   at DiffSharp.Backend.OpenBLAS.BLAS.dnrm2_(Int32* n, Double* x, Int32* incx)
   at DiffSharp.Backend.OpenBLAS.BLAS.dnrm2(Double[] x)
   at DiffSharp.AD.Float64.DV.L2Norm(DV a)
```

If I instead install with Homebrew:

```
brew install openblas
```

(which brew won't link globally "Warning: Refusing to link macOS-provided software: openblas")

then go ahead an link locally in a project

```
ln -s /usr/local/Cellar/openblas/0.3.6_1/lib/*.dylib .
```

I get everything working. So the choice seems to be to package more of these files or rely on the end-user.

@erydo I saw that you did some work on this, any suggestions on the best approach?

I'm going to default to leaving it in the end-users responsibility. That's what the last released NuGet did.

cgravill:
Merging this to provide a test release.

PR #56, opened 2019-06-25
Add set-up of Mul_DMCons_DM for adjoint calculation

cgravill:
Fixes an initialisation issue with constant matrix * matrix

Added a simple gradient descent test, perhaps need to separate out these smoke tests or start controlling the randomness and check results.


cgravill:
Note there may be other reachable cases that have an equivalent issue. We might need a different strategy to ensure all the right cases are handled:

https://github.com/DiffSharp/DiffSharp/blob/0fb1a8cec3e0706d792c79d18c3650fc48a3711f/src/DiffSharp/AD.Float32.fs#L3471-L3569

cgravill:
Spoke with @dsyme directly I'm going to merge this specific fix.

I'll take a look at a more general change to separate out the scalar, vector, and matrix operations separately.

Issue #57, opened 2019-07-07
Support more target frameworks in DiffSharp.fsproj


allisterb:
Right now only the netstandard2.0 target framework is supported by the DiffSharp project file. This causes build failures in libraries that use DiffSharp and must support net45 and older .NET frameworks and toolchains. Both NET Standard 2.0 and other targets can be supported by using this line in DiffSharp.fsroj

`<TargetFrameworks>netstandard2.0;netcoreapp2.0;net45</TargetFrameworks>`

I built DiffSharp successfully for these targets without any changes to the source code and the tooling will take care of sorting out the dependencies and packaging the libraries correctly for NuGet.

dsyme:
I think it's better for the project if it focuses on moving ahead and assumes .NET Standard 2.0 (or indeed .NET Standard 2.1 if it proves necessary)

Is there any reason you can't update to .NET 4.7.2, which supports .NET Standard 2.0 components?

allisterb:
The main reason is I'd like to run my F# projects on Azure Notebooks which uses Mono 5.1 and I thought I should target the lowest .NET framework possible for safety and to minimize problems. I don't know if NuGet on Mono 5 can restore .NET Standard packages: https://stackoverflow.com/questions/53479603/reference-netstandard-nuget-package-in-4-7-2-project-mono

dsyme:
@allisterb I see.  Yes, that F# implementation in Azure Notebooks should really be updated....

dsyme:
@allisterb The thing is, we expect we will very soon get other .NET Standard 2.0 dependencies, e.g. if/when tensor support is added.

dsyme:
Until then I guess we could add `net45` but please assume it's going away at some point...

allisterb:
Actually I could just make my own fork and maintain the net45 target there. I may have to make a fork in any case because I'd like to integrate DiffSharp with my project and I might need access to the internals. So don't bother to add net45 unless there's more demand for it

dsyme:
> Actually I could just make my own fork and maintain the net45 target there. I may have to make a fork in any case because I'd like to integrate DiffSharp with my project and I might need access to the internals. So don't bother to add net45 unless there's more demand for it

I think it's ok to add it for now, since it is easy and fits with the 0.8 release series.  But that might be the last series supporting it!

allisterb:
ok cool, thanks!

dsyme:
@allisterb COuld you send a PR for this?  thanks

[...]

PR #58, opened 2019-07-11
Add net45 target framework to DiffSharp.fsproj.

allisterb:
#57


allisterb:
> The reference assemblies for framework ".NETFramework,Version=v4.5" were not found. To resolve this, install the SDK or Targeting Pack for this framework version or retarget your application to a version of the framework for which you have the SDK or Targeting Pack installed.

 The Travis CI server needs the .NET 4.5 targeting pack: https://docs.microsoft.com/en-us/dotnet/framework/install/guide-for-developers

cgravill:
The build is done through Docker, so it might just be a case of adding Mono to that:
https://github.com/DiffSharp/DiffSharp/blob/master/Dockerfile

https://www.mono-project.com/download/stable/#download-lin

PR #59, opened 2019-08-15
Correction to matrix slicing for start point of AddSubMatrix

cgravill:
I've traced through an issue we hit with array slicing. I believe it's supplying "row finish" rather than "column start" which doesn't match how it's used:

https://github.com/DiffSharp/DiffSharp/blob/855b1926d422744ac38d8750c57957e675dc2046/src/DiffSharp/AD.Float32.fs#L2358-L2359

Smoke test demonstrating index out of bounds.


cgravill:
I did some more checks and review with @byoyo and this looks good. I'll go ahead with the merge and release of a test version

PR #63, opened 2019-11-05
Code formatting - add space line between all members

dsyme:
This allows `///` comments to be inserted later and is pretty much normal for all F# code


gbaydin:
Thank you

PR #65, opened 2019-11-05
Split out reference implementation of tensors

dsyme:
Not yet finished


dsyme:
This is done

gbaydin:
Thank you very much @dsyme !

kevmal:
Maybe there should be a device id for multi gpu/cpu support? I could get a MXNet backend going pretty quick but perhaps it's too early?

kevmal:
@dsyme @gbaydin I went ahead and prototyped an MXNet backend:
https://github.com/kevmal/DiffSharp/blob/mxnet/src/DiffSharp.Backend.MXNet/RawTensorFloat32CPU.fs

All tests pass. Though it's specifically `RawTensorFloat32CPU` the implementation easily generalizes to whatever.

PR #71, opened 2020-02-14
Simplify RawTensor API and RawTensorFloat32CPU

dsyme:
@gbaydin This ssimplifies the `RawTensorFloat32CPU` so you don't get that weirdness of going through the statics thunk thing jsut for simple `RawTensorFloat32CPU` coding. 

There are now also far fewer casts needed in `RawTensorFloat32CPU.fs`

### DIFF
--- a/src/DiffSharp.Backend.None/RawTensorFloat32CPU.fs
+++ b/src/DiffSharp.Backend.None/RawTensorFloat32CPU.fs
@@ -443,13 +443,13 @@ type RawTensorFloat32CPU(values: float32[], shape:int[]) =
         let result = t.Values |> Array.map abs
         upcast RawTensorFloat32CPU(result, t.Shape)
 
-    override t1.ReluT() =
-        let result = t1.Values |> Array.map (max 0.f) 
-        upcast RawTensorFloat32CPU(result, t1.Shape)
+    override t.ReluT() =
+        let result = t.Values |> Array.map (max 0.f) 
+        upcast RawTensorFloat32CPU(result, t.Shape)
 
-    override t1.SigmoidT() =
-        let result = t1.Values |> Array.map (fun v -> 1.f / (1.f + exp -v))
-        upcast RawTensorFloat32CPU(result, t1.Shape)
+    override t.SigmoidT() =
+        let result = t.Values |> Array.map (fun v -> 1.f / (1.f + exp -v))
+        upcast RawTensorFloat32CPU(result, t.Shape)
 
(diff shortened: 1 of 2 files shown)


gbaydin:
Thank you. Everything good exceptI think I had some good reason to have a generic `Value:obj` field in `RawTensor` which had to do with future backends, but let's go with the design in this PR for the time being.

gbaydin:
Thank you. Everything good exceptI think I had some good reason to have a generic Value:obj field in RawTensor which had to do with future backends, but let's go with the design in this PR for the time being.

gbaydin:
Thank you. Everything good except I think I had some good reason to have a generic `Value:obj` field in `RawTensor` which had to do with future backends, but let's go with the design in this PR for the time being.

dsyme:
> Thank you. Everything good except I think I had some good reason to have a generic Value:obj field in RawTensor which had to do with future backends, but let's go with the design in this PR for the time being.

I expect the GPU subtypes will define their own field.

dsyme:
Separately, I'm beginning to wonder if we should be using SciSharp's NumPy.NET as the RawTensor type.   It feels like we are re-inventing the wheel for this part. https://github.com/SciSharp/Numpy.NET

gbaydin:
So https://github.com/SciSharp/Numpy.NET is a .net binding for Python numpy, running Python in the background http://pythonnet.github.io/
There is also https://github.com/SciSharp/NumSharp that is a port of numpy to C#. I think a `RawTensor` backend can be based on one of these when the time comes. Perhaps the C# one would be better because there is no Python running.

Independent from that, the work I'm doing with the pure F# "None" backend in `RawTensorFloat32CPU` is needed for me to understand the internals of elementary numeric operations in order to design the nested derivative rules correctly, and also have a correctness reference for unit testing any new backends when they show up later.

dsyme:
> There is also https://github.com/SciSharp/NumSharp that is a port of numpy to C#. 

Yes I meant NumSharp, my apologies.

> I think a RawTensor backend can be based on one of these when the time comes. 

Hmmm I hadn't thought of that - yes, you're right that may be possible.  Interesting.

> Independent from that, the work I'm doing with the pure F# "None" backend in RawTensorFloat32CPU is needed for me to understand the internals of elementary numeric operations in order to design the nested derivative rules correctly, and also have a correctness reference for unit testing any new backends when they show up later.

Yes, it is really incredibly nice to have a completely detached implementation.

PR #72, opened 2020-02-15
Enable unused variable warnings

dsyme:
This enables unused variable warnings.  This consistently results in fewer bugs so we should enable it now

@gbaydin The variable `is` is an unused variable here, this is a bug or a mistake

```fsharp
            let is:int[] = value.ToArray() :?> obj[] |> Array.map toInt
            Seq.init d.BatchShape.[0] (fun i -> d.Probs.[i]) |> Tensor.Stack |> Tensor.Log
```


dsyme:
> Line by line review found only one potential oversight with Test/Program.fs line 87. The unused value was previously used in a print out and not for the side effect. It might be worth commenting out this Tensor.Create as well or commenting the corresponding print statement.

Thanks, fixed!

gbaydin:
`Test/Program.fs` is just scratch space for my debugging and test-writing efforts. Please do not waste your time with that part.

PR #73, opened 2020-02-15
None/CPU backend - Float64, Int32 tensors

dsyme:
@gbaydin Implements raw tensors of various types for `DiffSharp.Backend.None`

There is no real hurry to merge this, I can maintain it for a while as we talk about it. 

### Approach 

1. Backend types `RawTensorFloat32CPU`, `RawTensorFloat64CPU`  and  `RawTensorInt32CPU` all inherit from `RawTensorCPU`

2. The non-math code for raw CPU tensors moves to `RawTensorCPU`

3. The math code for raw CPU tensors moves to a module `RawTensorCPU` using generic inline code.

4. RawTensorFloat32CPU, RawTensorFloat64CPU  and  RawTensorInt32CPU are all relatively short files bringing the base functionality through to the specific types.

5. Operations like "t.Sin()" on Int32 tensors first convert to a Float32 tensor and implicitly return a Float32 tensor

Pros:

1. Complete the matrix for Backend.None
2. bug-for-bug consistency across the CPU tensor types
3. the interactions betwewen the CPU tensor types can be dealt with in one place consistently
4. performance should be no worse than before since the math code is inlined.
5. perf improvements apply across all RawTensorCPU types

Cons:

1. Adding an operation requires writing a bit of code in each of RawTensorFloat32CPU, RawTensorFloat64CPU, RawTensorInt32CPU. However you always have the option of prototype only with Float32 and do the generalization later.

2. Writing the generic inline math code can be tricky, though I've done my best ot make it as simple as possible.

3. Expands out the matrix of combinations considerably.


dsyme:
There's a design issue here for `tensor.Cast(...)`.  See this code:
```fsharp
    member t.Cast(dtype) =
        match t with
        | Tensor(tp) -> Tensor(tp.Cast(dtype))
        | TensorF(tp,td,tag) -> TensorF(tp.Cast(dtype), td.Cast(dtype),tag)
        | TensorR(tp,_,_,_,_) -> failwith "cannot cast TensorR nodes"
```
The questions are

1. Do we support a `t.Cast(...)` operation at all? 
2. Does it need to work on F and R nodes?
3. If it needs to work on F nodes, do we expect the Primal and Derivative both to be cast?
4. If it needs to work on R nodes, what's the spec?

I assume the answers are 
1. Yes
2. No
3. n/a
4. n/a

dsyme:
Note it is `t.double()` , `t.single()` etc. and `t.type(...)` in PyTorch

PR #75, opened 2020-02-15
Feature: keepDim parameter

dsyme:
Reopen #70 so @gbaydin can review properly


gbaydin:
Thank you. I think we would also need tests for the derivatives, because the derivative code perhaps need to be changed to handle `keepDim`. Also tests for `Variance` would be good.

gbaydin:
In general, I test these operations in PyTorch to convince myself about api compatibility and numerical correctness. It would be great if you could share a bit of Python code that replicates these cases.

dsyme:
I think we can pull this now.

* [x] added testing for derivatives and comparisons with  PyTorch.   
* [x] added some testing in TestTensor.fs for cases not currently covered (e.g. variance)
* [x] the Python code fragments are in inline comments in the TestTensor.fs and TestDerivatives.fs, e.g. see https://github.com/DiffSharp/DiffSharp/pull/75/files#diff-79888ca94c742a41ec58bcc44dea5c97R826.   I find that handy to keep as a record - and maybe in the future we can automate these parts by actually running Python in the test DLL

p.s. There was a transient failure in the test `TestDistributionsUniformBatched` - I think there is something slightly awry with that test as I've seen it fail once before.

gbaydin:
> p.s. There was a transient failure in the test `TestDistributionsUniformBatched` - I think there is something slightly awry with that test as I've seen it fail once before.

This sounds OK. Those tests are stochastic and involve sampling from probability distributions. I will revisit them soon and the tolerances can be adjusted to minimize failing tests.

gbaydin:
Thank you, I reviewed these and everything looks good.

PR #76, opened 2020-02-15
Unused variable warnings

dsyme:
This enables unused variable warnings. This consistently results in fewer bugs so we should enable it now

@gbaydin The variable is is an unused variable here, this is a bug or a mistake

            let is:int[] = value.ToArray() :?> obj[] |> Array.map toInt
            Seq.init d.BatchShape.[0] (fun i -> d.Probs.[i]) |> Tensor.Stack |> Tensor.Log

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -1,0 +1,7 @@
+<Project>
+
+  <PropertyGroup>
+    <OtherFlags>--warnon:1182 $(OtherFlags)</OtherFlags>
+  </PropertyGroup>
+
+</Project>
--- a/src/DiffSharp.Core/Tensor.fs
+++ b/src/DiffSharp.Core/Tensor.fs
@@ -2,6 +2,8 @@
 open DiffSharp.Backend
 open DiffSharp.Util
 
+#nowarn "1182" // turn off compiler-generated unused variable warnings in this file only
+
 [<CustomEquality; CustomComparison>]
 type Tensor = 
     | Tensor of RawTensor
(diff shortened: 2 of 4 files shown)


src/DiffSharp.Core/Tensor.fs
```diff
@@ -222,55 +222,55 @@ type Tensor =
         if a.Shape = b.Shape then
             let inline fRaw(a:RawTensor,b) = a.AddTT(b)
             let inline fTensor(a,b) = a + b
-            let inline dfTensorFwdTT(cp,ap,ad,bp,bd) = ad + bd
-            let inline dfTensorFwdTC(cp,ap,ad) = ad
-            let inline dfTensorFwdCT(cp,bp,bd) = bd
+            let inline dfTensorFwdTT(_cp,_ap,ad,_bp,bd) = ad + bd
```
dsyme: @gbaydin These changes are needed to suppress the unused variable warnings.  I think they are clarifying as they capture the intent that various parts of the derivative depend on the inputs, and you have to explicitly mark up when they don't.

src/DiffSharp.Tests/TestDerivatives.fs
```diff
@@ -235,7 +235,6 @@ type TestDerivatives () =
             [7.5281; 1.4816];
             [5.0328; 2.2756]]).ReverseDiff()
         let revz = Tensor.MatMul(revx, revy)
-        let revzCorrect = Tensor.Create([0.2; 0.333333; 0.428571])
```
dsyme: @gbaydin This is an example of the kind of thing this warning captures

Directory.Build.props
```diff
@@ -0,0 +1,7 @@
+<Project>
+
+  <PropertyGroup>
+    <OtherFlags>--warnon:1182 $(OtherFlags)</OtherFlags>
```
dsyme: @gbaydin This turns on the warning for all projects in the repository.  You can put other build settings in this file and they will also apply across the whole repository, it;s a new .NET SDK thing

dsyme:
I talked this ovder with @gbaydin.  He groks that there's value to having compiler-generated unused variable warnings for the project overall but would prefer the unused variable warning is not activated in Tensor.fs as he already gets good warnings from the IDE tooling  in VSCode

So I removed the changes in Tensor.fs and added a `#nowarn` for now

PR #77, opened 2020-02-16
Feature: Backend.None typed tensors

dsyme:
(Replaces #73, using a feature branch)

The plan is that DiffSharp 1.0 will support the usual PyTorch range of typed tensors.
This implements typed tensors for `DiffSharp.Backend.None`. 

There is no real hurry to merge this, I can maintain it as `dev` progresses.

Backend.None types share code in `RawTensorCPU<T>`. The non-math code for raw CPU tensors moves to `RawTensorCPU`. The math code for is in a module `RawTensorCPU` using generic inline code. Specific types inherit and incorporate the code needed for construction

For Int32, operations like "t.Sin()" on Int32 tensors first convert to a Float32 tensor and implicitly return a Float32 tensor, though we could make them just fail.

Testing is expanded out to test the full range of types.

Pros:

1. Complete the matrix for Backend.None
2. Bug-for-bug consistency across the CPU tensor types
3. The interactions betwewen the CPU tensor types can be dealt with in one place consistently
4. Performance should be no worse than before since the math code is inlined.
5. Perf improvements apply across all RawTensorCPU types

Cons:

1. Adding an operation requires writing a line of code for each of RawTensorFloat32CPU, RawTensorFloat64CPU, RawTensorInt32CPU. You can prototype only with Float32 and do the generalize later.

2. Writing the generic inline math code can be tricky, though I've done my best to make it simple.


dsyme:
I need to make an adjustment to this if we want to align to PyTorch.

The PyTorch behaviour for default dtype is as follows:
* `torch.tensor([True, False]).dtype` --> bool
* `torch.tensor([1, 2]).dtype` --> int64
* `torch.tensor([1.0, 2.0]).dtype` --> float32
* `torch.tensor([1, 2.0]).dtype` --> float32
* `torch.tensor([3.40282347e39])` --> float32 (despite being over....)

dsyme:
@gbaydin I updated this branch to integrate with `dev`

dsyme:
@gbaydin I'm working to improve code coverage on this.

The main issue is that the PR adds inlined F# math code, which doesn't get tracked by code coverage tooling.   Do you know any workaround for that?

gbaydin:
> @gbaydin I'm working to improve code coverage on this.

I didn't mean to impose such strict coverage requirements at this point. It's basically codecov's default settings that I didn't tune yet.

Coverage reports are based on https://github.com/tonerdo/coverlet (referenced here https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Tests/DiffSharp.Tests.fsproj#L11) but I have a feeling that it doesn't do a good job with handling F# features like generic code, `inline` etc. I haven't checked closely yet.

dsyme:
> I didn't mean to impose such strict coverage requirements at this point. It's basically codecov's default settings that I didn't tune yet.

It's pretty handy and forced me to go back assess the testing in useful ways.  It's the first time I've worked on a project with it enabled, it adds an interesting extra dimension.

> Coverage reports are based on https://github.com/tonerdo/coverlet (referenced here https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Tests/DiffSharp.Tests.fsproj#L11) but I have a feeling that it doesn't do a good job with handling F# features like generic code, inline etc. I haven't checked closely yet.

You can add ExcludeFromCodeCoverage attribute, as you've already noticed.  Yes, there's no other way to deal with inline code.

dsyme:
>  there's no other way to deal with inline code.

Actually there is a way to get coverage for inline code, which is to remove the attribute and do a quotation-based dynamic invoke of the code.  However it won't work for some generic math code until https://github.com/dotnet/fsharp/pull/6810 is fixed

dsyme:
Am working through the code coverage and comparing torch. 

In addition to [this]() above, there are some other things we need to decide here

1. Type promotion.  Torch does some type promotion, e.g. adding 'int64' and 'float' tensors

2. Generation of 'bool' tensors.  In Torch operations like `<` produce boolean tensors, e.g.

       t = torch.tensor([[1]], dtype=torch.bool)
       t.lt(t).dtype

   gives dtype `bool`.  We would have to adjust the library for this.

3. Do we really want to exclude `Tensor<T>`?  The Torch and (current) DiffSharp 1.0 designs have loosely typed tensors `Tensor` not `Tensor<int64>`.  If we support type promotion we will *never* really be able to have strongly element-typed tensors, and will have to rely on live checking tooling etc.  (Well, it may be possible to make a wrapper API I guess, restricting the operations and requiring more casts)

gbaydin:
> 1. Type promotion.  Torch does some type promotion, e.g. adding 'int64' and 'float' tensors

I think we should also do this, but I don't have a particularly strong opinion on this currently.

> 2. Generation of 'bool' tensors.  In Torch operations like `<` produce boolean tensors, e.g.

Yes, I think this should be the behavior in DiffSharp. Actually it was the main reason I faced the need to have boolean tensors recently. Currently we have `1.`s and `0.`s in float tensors representing boolean values.

> 3. Do we really want to exclude `Tensor<T>`?  The Torch and (current) DiffSharp 1.0 designs have loosely typed tensors `Tensor` not `Tensor<int64>`.  If we support type promotion we will _never_ really be able to have strongly element-typed tensors, and will have to rely on live checking tooling etc.  (Well, it may be possible to make a wrapper API I guess, restricting the operations and requiring more casts)

I think avoiding `Tensor<T>` is good as it simplifies the design and the user-facing code. Personally as a machine learning user I would feel much more comfortable and "at home" with just a `Tensor` type. In practice people just use a default single-precision `Tensor` type and perhaps sometimes go to half-precision. And you move them to GPU by converting things with a `.cuda()` call on instances or some similar mechanism. Also we need to keep in mind that types not available in dotnet can be needed in practice (such as the half-precision type that is not currently available but can be a part of net 5.0 https://github.com/dotnet/runtime/issues/936).

dsyme:
>I think we should also do this, but I don't have a particularly strong opinion on this currently.

Cool, agreed on all this, I've updated the PR for this including testing

> Generation of 'bool' tensors. In Torch operations like < produce boolean tensors, e.g. Yes, I think this should be the behavior in DiffSharp. Actually it was the main reason I faced the need to have boolean tensors recently. Currently we have 1.s and 0.s in float tensors representing boolean values.

This is now in the PR too

dsyme:
@gbaydin This is ready now

dsyme:
Great!  So glad to have this in!

PR #78, opened 2020-02-16
Backend.None: Adjust code order to minimize later diffs

dsyme:
This is a cosmetic rename and code-order change in Backend.None to make the later diff of features like #77 much smaller.

Basically the operations in `RawTensorFloat32CPU` are placed in order so the "structural" (non-math) operations go first, and the math operations second.

The rename `RawTensorFloat32CPU.fs`  → `RawTensorCPU.fs` is also done again to minimize later diffs (rather than have #77 split up the file).


gbaydin:
Thank you

PR #80, opened 2020-02-17
test conv1d with batchsize 1

kevmal:
Adds an additional check on conv1d.  Conv1d with stride > 1 assumes `GetSlice` only drops one dim.


gbaydin:
Thank you for noticing this! The code is now fixed to handle this correctly.

PR #82, opened 2020-02-18
Explain the meaning of tensor tuple fields

pkese:
Annotated tensor tuples with descriptive names to help newbies (like me) understand what is going on.


pkese:
Or maybe `tensor` should better be named `primal` to match the meaning of `t.Primal`.
Anyways, it is just a suggestion.

gbaydin:
Thank you. I'm merging this and I will subsequently update the names to be consistent with the rest of the code.

PR #84, opened 2020-02-24
Testing+fix for tensor comparison and hashing

dsyme:
This fixes some things with tensor comparison and hashing

* `CompareTo` was only implemented for scalars.  This implements it for all `RawTensor`

* In F#, the behaviour of `Equals` and `GetHashCode` should align so that two equal tensors give equal hashcodes.   The current `tensor.GetHashCode()` didn't follow this rule as it hashed a `TensorF` differently to a `Tensor` or `TensorR`.  I propose we make `GetHashCode` follow the same rule as `CompareTo` and `Equals` where only the `PrimalRaw` is considered for the hash.  I think this is the only spec that really makes sense.

Also adds much more systematic testing for hashing, comparison and equality on `Tensor`.


gbaydin:
So we've been reviewing this @dsyme, and I think everything is fine except we shouldn't implement comparison (of the type expressed by `System.IComparable`) between tensors. Operations that return element-wise comparison results like `Tensor.Ge` and `Tensor.Le` should be used instead.

Issue #69, opened 2020-02-14
Roadmap to DiffSharp 1.0


dsyme:
This is a catalog issue to track what needs doing for DiffSharp 1.0, based on 1:1 discussions with @gbaydin.  There will be a long list of other things, we'll extend this as necessary

* [x] Typed Backend.None CPU tensors ([draft](https://github.com/DiffSharp/DiffSharp/pull/77))

* [x] Add `keep_dims` on `Mean` ([done](https://github.com/DiffSharp/DiffSharp/pull/70))

* [x] Fix `CompareTo` in RawTensorFloat32CPU.fs ([done](https://github.com/DiffSharp/DiffSharp/pull/84))

* [x] Broadcasting. Full pytorch-style broadcasting for `Add` (see [TODO here](https://github.com/DiffSharp/DiffSharp/blob/9aa23cff72b8b8eb2cb0ebbf2973a8d892ad186b/src/DiffSharp.Core/Tensor.fs#L278)).   The design principle expected here is "we should do the same thing as PyTorch".  Similarly full pytorch-style broadcasting for `Mul` and other operations. ([done](https://github.com/DiffSharp/DiffSharp/pull/85))
 
* [x] Convolutions (@gbaydin)

* [x] Transposed convolutions

* [x] Batchnorm

* [x] Dropout

* [x] Switch to Python-style casing on all operations to align with SciSharp

* [x] Remove excess overloads and use optional arguments instead

* [x] Finalize API `dsharp.abc`

* [x] libtorch and cuda backends

* [x] Add `Reshape` (similar code to `View`)

* [x] Add `OneHot`

* [x] Differentiation API

* [x] Optimizers

* [x] Fix possible memory leak on Linux

* [x] Tensor save/load

* [x] General transpose 

* [x] probability distributions

* [x] Generalization and batching of `Transpose`.

* [x] Zero-size tensors #150

* [x] Docs tooling #134

* [x] Docs #167

* [x] PyTorch Half support

* [x] Batching conv1d/2d/3d/... https://github.com/DiffSharp/DiffSharp/issues/98

* [x] Batching for `MatMul`.  Currently [no batching is supported, only 2D x 2D](https://github.com/DiffSharp/DiffSharp/blob/9aa23cff72b8b8eb2cb0ebbf2973a8d892ad186b/src/DiffSharp.Core/Tensor.fs#L434) [done](https://github.com/DiffSharp/DiffSharp/pull/88)

[...]

PR #85, opened 2020-02-24
Feature: expand, plus broadcasting for +/-/*/divide

dsyme:
Implement broadcasting

* [x] changes `tensor.Extend(...)` into a full `tensor.Expand(...)` that broadcasts through prefix of dimensions plus expanding `1` --> `N`.

* [x] implement broadcasting for `+`, `-`, `*`, `/`

* [x] add testing for `Expand`

* [x] add testing for broadcasting for `+`, `-`, `*`, `/`

* [x] add testing for derivatives of broadcasting for `+`, `-`, `*`, `/`

I did additional tested by commenting out the existing Scalar/Vector/Matrix extension.  Will discuss with @gbaydin on Friday

As an aside, to match PyTorch the `Expand` should use a view with a stride=0 in the appropriate dimensions (and thus not reallocate anything), but this requires strided views.  I'll add a TBD about that in #69


gbaydin:
Thank you! The derivative tests are a bit hard to follow. Can we please improve those. I will look through the code and check everything carefully as well.

dsyme:
I've added the explicit tests for the derivatives of `expand` here https://github.com/DiffSharp/DiffSharp/pull/85/files#diff-79888ca94c742a41ec58bcc44dea5c97R73

I think this is now ready.  Would appreciate python code to test the forward derivatives

gbaydin:
Thank you. I reviewed these changes and I'm convinced they are fine pending some changes in testing that I will implement.

PR #87, opened 2020-02-25
fix #86

### LINKED ISSUE #86: Failure in tupled tensor creation
Tupled tensor creation fails if the leaves are size-1 lists, e.g.

        let v = Tensor.Create [[2.], [3.], [4.]]

I'll send the fix shortly

dsyme:
Fixes this issue with tensor creation using tupled syntax:  https://github.com/DiffSharp/DiffSharp/issues/86


dsyme:
@gbaydin I will merge this as it's a simple fix to what you merged the other day

PR #88, opened 2020-02-26
Feature: Batch matmul

dsyme:
Adjust MatMul to support batching and broadcasting, in both Tensor and the reference implementation

Builds on #85


src/DiffSharp.Core/Tensor.fs
```diff
@@ -567,7 +577,7 @@ type Tensor =
     member t.SumT2Dim0() = Tensor.SumT2Dim0(t)
     
     static member Transpose (a:Tensor) =
-        if a.Dim <> 2 then failwithf "Expecting a 2d Tensor, received Tensor with shape %A" a.Shape
+        if a.Dim > 2 then failwithf "Expecting at least a 2d tensor, received Tensor with shape %A" a.Shape
```
gbaydin: Are we sure this is correct?

dsyme: No it's not correct, thanks!  Will make sure we test Transpose directly on batching

dsyme:
@gbaydin Can we get this in do you think?

dsyme:
I've merged this with `dev` and it should now be ready (once tests and coverage pass)

dsyme:
@gbaydin I've merged this with dev, it should now be ready

gbaydin:
I think the behavior of the 'transpose' operation in this branch is not consistent with PyTorch.

PyTorch transpose behavior is as follows:
```python
a = torch.randn([])
b = torch.randn([10])
c = torch.randn([10,20])
d = torch.randn([10,20,30])
e = torch.randn([10,20,30,40])
at = torch.t(a) # Gives shape []
bt = torch.t(b) # Gives shape [10]
ct = torch.t(c) # Gives shape [20, 10]
dt = torch.t(d) # Fails because d.dim > 2
et = torch.t(e) # Fails because d.dim > 2
```

DiffSharp behavior in this branch is as follows:

```fsharp
let a = dsharp.randn([])
let b = dsharp.randn([10])
let c = dsharp.randn([10;20])
let d = dsharp.randn([10;20;30])
let e = dsharp.randn([10;20;30;40])
let at = a.transpose() // Fails because a.dim < 2
let bt = b.transpose() // Fails because b.dim < 2
let ct = c.transpose() // Gives shape [20; 10]
let dt = d.transpose() // Gives shape [10; 30; 20]
let et = e.transpose() // Gives shape [10; 20; 40; 30]
```

I believe there is no need for a "batch-transpose" operation in general and batch transposition can be achieved when we have the general transpose operation `torch.transpose(input,dim0,dim1)` which can freely swap any two dimensions in all shapes of tensors. https://pytorch.org/docs/stable/torch.html#torch.transpose

I think this batch-transpose behavior was needed mainly for the reverse mode of batch matrix multiplication. We can either keep this and rename it to something like `batchTranspose` (and make it internal?) or implement general transpose (which is a needed operation anyway) and use it for this purpose.

dsyme:
Yup agreed. There's also `torch.permute` which is even more general right?

https://stackoverflow.com/questions/57512113/how-to-take-a-transpose-for-each-matrix-in-a-batch-in-pytorch

dsyme:
> keep this and rename it to something like batchTranspose (and make it internal?) or

This might be the simplest option for now, to get the batch matmul implementation and tests consolidated?  Then deal with the second issue?

gbaydin:
> > keep this and rename it to something like batchTranspose (and make it internal?) or
> 
> This might be the simplest option for now, to get the batch matmul implementation and tests consolidated? Then deal with the second issue?

I agree, it makes sense.

dsyme:
@gbaydin Can we relax the code coverage settings?  It's really tricky to get them going up all the time " 63.49% of diff hit (target 63.67%)"

dsyme:
@gbaydin This is now updated with `batchTranspose` internal

gbaydin:
> @gbaydin Can we relax the code coverage settings? It's really tricky to get them going up all the time " 63.49% of diff hit (target 63.67%)"

These are the default settings and I agree they are too strict. I looked into relaxing them before but it was surprisingly difficult to find the information. Let me look a bit better and fix it. :)

dsyme:
> These are the default settings and I agree they are too strict. I looked into relaxing them before but it was surprisingly difficult to find the information. Let me look a bit better and fix it. :)

If you like give me admin rights?  I can poke around

gbaydin:
> > These are the default settings and I agree they are too strict. I looked into relaxing them before but it was surprisingly difficult to find the information. Let me look a bit better and fix it. :)
> 
> If you like give me admin rights? I can poke around

I think you already have it. Codecov uses GitHub permissions https://codecov.io/gh/DiffSharp/DiffSharp/ Please let me know if it doesn't work.

gbaydin:
I think we need to set the project/patch targets to custom values using https://github.com/DiffSharp/DiffSharp/blob/dev/codecov.yml. See here: https://docs.codecov.io/docs/commit-status

gbaydin:
Inspecting this further, I have the following concerns about the PR:

- The behavior of DiffSharp `matmul` is not the same with [`torch.matmul`](https://pytorch.org/docs/stable/torch.html#torch.bmm) which has special cases for handling 1d arguments (e.g., doing matrix-vector products for 2d-1d combinations, dot product for the 1d-1d combination), doing a normal matmul (2d-2d), and doing a batched matmul (`bmm`, see below) with broadcasting according to the dimensionality of the arguments.
- In torch there is a [`torch.bmm`](https://pytorch.org/docs/stable/torch.html#torch.bmm) that does batch-matmul only for 3d tensors. But DiffSharp `matmul` is not the same with this operation either because it works with >=2d tensors and supports broadcasting.

I think if we can get `matmul` to behave identical to `torch.matmul`, it would be great. Introducing a `bmm` that works with 3d-3d only would be straightforward as well.

dsyme:
> Inspecting this further, I have the following concerns about the PR:

Cool, makes sense

gbaydin:
Some relevant discussions here: https://github.com/pytorch/pytorch/issues/18027 and here: https://github.com/tensorflow/tensorflow/issues/5523

dsyme:
@gbaydin 

> I think if we can get matmul to behave identical to torch.matmul, it would be great. Introducing a bmm that works with 3d-3d only is a minor thing and should be straightforward.


I have made and tested this adjustment based on the Pytorch docs for `matmul` and this should now be ready

> Another thing that we need to think about is whether to introduce MatMulT2T1 or MatMulT1T1 (or DotTT) ...

I haven't done these parts.

gbaydin:
Thank you, I will have a look and merge soon.

gbaydin:
Finally merged this (with a lot of delay!)

dsyme:
Something went wrong with the overall diff for this in Tensor.fs, which is showing 6000+ line changes (nearly all whitespace)

I'll try to undo and force push a simpler diff

PR #90, opened 2020-03-02
Move conv rev diff ops out of main loop

dsyme:
@gbaydin I moved the reverse gradient computations for `conv1d` and `conv2d` out of the main reverse gradient algorithm  - after we do #89 we will move them to a separate file in any case.

I'll pull this as I'll do batching for `conv1d` as part of #88 and the change is otehrwise cosmetic - please adjust the names of the (internal) operations as you see fit.


dsyme:
(The conv ops already suppoirt batching but I think it's right to move the reverse parts out of the differentiation loop - they belong next to the main op)

PR #95, opened 2020-04-24
Alternative approach to typed tensors

dsyme:
This is an alternative approach to #77

The tests remain the same, it's the reference implementation of CPU typed tensors that changes.  

Basically we have only one type `RawTensorCPU<_>` which gets instantiated.  No further subclasses.  Then the generic math is done by internalised type switch (which the .NET Core handles efficeintly, not that performance is a huge factor here)

The advantage is that it's less confusing to write the generic math code when doing things like convolutions.  It's still got to be generic, of course, but it's just less confusing than when we have the extra subtypes.


dsyme:
@gbaydin This is so cool!

image

dsyme:
Closing as we're not going with this approach

PR #96, opened 2020-04-27
Convnets

gbaydin:
This is to keep track of the initial convnet implementations. The API is currently evolving. Need to add 3d convolutions at a later point.


gbaydin:
This was good progress for convnets and model definition and training API in general. I will handle pooling operations (e.g., maxpooling) in another PR. Merging this now.

PR #100, opened 2020-05-05
remove unnecessary "inline" 

dsyme:
I think we should remove these unnecessary `inline`

The F# compiler will inline these in nearly all cases, and in the cases where it doesn't the performance difference is going to be completely irrelevant in the face of tensor operations.


gbaydin:
I've been wondering about the same. Ok let's get rid of them.

PR #103, opened 2020-05-07
Testing: Multi-backend testing and test separation of backends

dsyme:
In preparation for working on the native backends, this addresses two things with our testing:

1. Our tests currently don't test across backends and devices.  We will soon need to test across these.  This systematically makes it easy to adjust our tests to work across multiple backends and devices.

2. An important thing is that the internals of DiffSharp don't create tensors for the default backend "by accident", thereby getting, say, a Torch GPU tensor colliding with a reference implementation CPU tensor.  This adds testing for this.

The loops for testing acress types, devices and backends is now like this:

```fsharp
        for combo in Combos.IntegralAndFloatingPoint do 
            let t1 = combo.tensor([1.; 2.; 3.; 5.])
            let t2 = combo.tensor([1.; 3.; 5.; 4.])
            let t1t2Ge = t1.ge(t2)
            let t1t2GeCorrect = combo.tensor([1.; 0.; 0.; 1.], dtype=DType.Bool)

            Assert.AreEqual(t1t2GeCorrect, t1t2Ge)
            Assert.AreEqual(DType.Bool, t1t2Ge.dtype)
```
Here `combo` is a `Combo` object (replacing `DTypeInfo` and `info`) that represents a variation that adjusts one or all of the default device, default dtype or default backend.

I know you renamed `info` to `ty` and here I rename it back to `combo` but that's because it's not just a default DType - it's a default DType, default Device and default Backend.  If you want it to be a shorter name it could be `cx` or something?

See this too:

```fsharp
module Combos =

    let backends = ...
    let devices = ...
    let makeCombos dtypes =
        [ for backend in backends do
          for device in devices do
          for dtype in dtypes do
          yield ComboInfo(backend, device, dtype) ]
    ....
    let IntegralAndFloatingPoint = makeCombos DTypes.IntegralAndFloatingPoint
    let All = makeCombos DTypes.All
    ....
```
The default setting at the moment is to test using both the default reference backend and a copy of it. This meand we would now get failures if our test or implementation code produces and compares tensors from different backends. It's really important to make sure backends don't mix inadvertently (it's one of the most pernicious problems in these implementations) which is why I'm adding the testing for this now.

```fsharp
    let backends = [ Backend.None; Backend.Register("TestDuplicate") ]

    let devices = [ Device.CPU ]
``

### CODE REVIEW

**Comment by dsyme** [comment]
@gbaydin The loops for testing acress types, devices and backends is now like this:

```
        for combo in Combos.IntegralAndFloatingPoint do 
            let t1 = combo.tensor([1.; 2.; 3.; 5.])
            let t2 = combo.tensor([1.; 3.; 5.; 4.])
            let t1t2Ge = t1.ge(t2)
            let t1t2GeCorrect = combo.tensor([1.; 0.; 0.; 1.], dtype=DType.Bool)

            Assert.AreEqual(t1t2GeCorrect, t1t2Ge)
            Assert.AreEqual(DType.Bool, t1t2Ge.dtype)
```
Here `combo` is a `Combo` object (replacing `DTypeInfo`) that represents a variation that adjusts one or all of the default device, default dtype or default backend.  

I know you renamed `info` to `ty` and here I rename it back to `combo` but that's because it's not just a default DType - it's a default DType, default Device and default Backend.


See this too:
```
module Combos =

    let backends = ...
    let devices = ...
    let makeCombos dtypes =
        [ for backend in backends do
          for device in devices do
          for dtype in dtypes do
          yield ComboInfo(backend, device, dtype) ]
```
The default setting at the moment is to test using **both** the default reference backend and a copy of it.  This meand we would now get failures if our test or implementation code produces and compares tensors from different backends.  It's really important to make sure backends don't mix inadvertently (it's one of the most pernicious problems in these implementations) which is why I'm adding the testing for this now.
```
    let backends = [ Backend.None; Backend.Register("TestDuplicate") ]

    let devices = [ Device.CPU ]
```

dsyme:
@gbaydin I've integrated this with dev - I think we should just pull it now, to avoid future conflicts.  The testing is sound and good

gbaydin:
Great, I will check it later today and merge very soon.

gbaydin:
I checked this and the testing is sound and good, as you said! Merging now.

PR #106, opened 2020-05-13
LibTorch backend (feature/torch2)

dsyme:
This is initial work on a LibTorch 1.0.1 backend using TorchSharp.

It requires a parallel build of TorchSharp using https://github.com/xamarin/TorchSharp/pull/127, as I've had to extend TorchSharp with numerous additional operations

The backend is not large and I am refactoring parts of it out.

### Notes

In pytorch there are two tensor types - Aten's tensor  for raw tensor values, and pytorch's Tensor for tensors that support gradients.  (this is the same as DiffSharp, which has RawTensor and Tensor for these).

So I am using

    DiffSharp Tensor --> DiffSharp RawTensor --> PyTorch tensor (but we don't ever set requires_grad on these).

After one false start I'm now just ignoring the existence Aten tensors, since there are lots of operations only implemented on PyTorch tensor.

### Dev Notes

Parallel clone of https://github.com/xamarin/TorchSharp/pull/127

To run TorchSHarp tests in debug:

    .\build &&  .\build -RunTests   

To build local TorchSharp packages:

    .\build -release &&  .\build -release -buildpackages 

To consume new packages into DiffSharp (may require version number update on package references) 

    rmdir /q /s \Users\dsyme\.nuget\packages\atensharp && rmdir /q /s \Users\dsyme\.nuget\packages\torchsharp && rmdir /q /s \Users\dsyme\.nuget\packages\libtorch.redist && devenv DiffSharp.sln

### Status 

* [x] Get the remaining operations implemented and tests passing
* [x] Get changes into TorchSharp and update the nuget packages for it
* [x] Update TorchSharp to more recent redist 
* [x] Check issues with potential memory leaks (see issue in TorchSharp repo)
* [x] Check TODOs
* [x] Enable testing for other types


dsyme:
Progress! 

   203 Passed
   23 Failed

Basically just convolutions, save/load, some distributions and associated operations to go!

dsyme:
Good news!  **This PR now fully passes all DiffSharp tests** when using TorchLib as the tensor implementation.

My next goal is to get the changes into TorchSharp, and publish those packages (I have permissions to publish).

gbaydin:
> In pytorch there are two tensor types - Aten's tensor for raw tensor values, and pytorch's Tensor for tensors that support gradients. (this is the same as DiffSharp, which has RawTensor and Tensor for these).
> So I am using
> ```
> DiffSharp Tensor --> DiffSharp RawTensor --> PyTorch tensor (but we don't ever set requires_grad on these).
> ```
Yes, I agree using PyTorch tensors with requires_grad=False is a good choice.

> After one false start I'm now just ignoring the existence Aten tensors, since there are lots of operations only implemented on PyTorch tensor.

This page gives a very nice summary of the PyTorch C++ structure. I think the level we can interface with is **C++ Frontend** in that list, which implements operations that closely mirror the PyTorch operations in Python. I think this is already what you've been doing. See "Closely model the Python frontend in its design" in [this page](https://pytorch.org/cppdocs/frontend.html#philosophy).

gbaydin:
> Good news! **This PR now fully passes all DiffSharp tests** when using TorchLib as the tensor implementation
> 
> (NOTE, this PR will still be red until the necessary packages are available).
> 
> My next goal is to get the changes into TorchSharp, and publish those packages (I have permissions to publish).

Sounds exciting!

dsyme:
> This page gives a very nice summary of the PyTorch C++ structure. I think the level we can interface with is C++ Frontend in that list, which implements operations that closely mirror the PyTorch operations in Python. I think this is already what you've been doing. See "Closely model the Python frontend in its design" in this page.

Yes, this is the level that TorchSharp exposes.  

Matt is working to codegen more or less the entire API TorchSharp API from Python metadata, or at least the functions we need.  There is also a Rust project that does this that we can learn from.

dsyme:
@gbaydin This PR contains quite a bit of refactoring and code-sharing to make writing a backend easier, and also additional testing, including tweaking the reference implementation and tests here and there in minor ways to match even torch more closely.  So I recommend holding off from any attempt to do a different backend until this is in.

gbaydin:
> @gbaydin This PR contains quite a bit of refactoring and code-sharing to make writing a backend easier, and also additional testing, including tweaking the reference implementation and tests here and there in minor ways to match even torch more closely. So I recommend holding off from any attempt to do a different backend until this is in.

Understood, thank you for the heads up.

dsyme:
This branch is now updated to Torch 1.5.0 .  The corresponding TorchSharp support is going through CI at the moment here: https://github.com/xamarin/TorchSharp/pull/137

After that goes through we should be able to get this green

src/DiffSharp.Core/Util.fs
```diff
@@ -121,7 +278,7 @@ let inline arraysAllClose (relativeTolerance:'T) (absoluteTolerance:'T) (array1:
     let dim1 = array1.Length
     let dim2 = array2.Length
     if dim1 <> dim2 then false
-    else Array.map2 (fun a b -> abs(a-b) <= absoluteTolerance + relativeTolerance*abs(b)) array1 array2 |> Array.forall id
+    else Array.map2 (fun a b -> abs(a-b) < absoluteTolerance + relativeTolerance*abs(b)) array1 array2 |> Array.forall id
```
gbaydin, line 281: The `<=` that I see replaced in this commit with `<` ultimately comes from https://numpy.org/doc/1.18/reference/generated/numpy.isclose.html and https://docs.python.org/dev/library/math.html#math.isclose that PyTorch also inherits.

It would be good to keep it as `<=`

gbaydin:
I think `xamarin/TorchSharp` currently has Mac builds disabled https://github.com/xamarin/TorchSharp/issues/135. Is my understanding correct that it will be straightforward to fix once the CI settings are updated?

src/DiffSharp.Backends.Reference/Reference.RawTensor.fs
```diff
@@ -0,0 +1,1550 @@
...
+    let inline Random ofDouble (shape:int[]) : (^T[] * int[]) =
+        let values = Array.init (shapeLength shape) (fun _ -> ofDouble (DiffSharp.Util.Random.Uniform()))
+        (values, shape)
+
+    let inline RandomNormal ofDouble (shape:int[]) : (^T[] * int[]) =
+        let values = Array.init (shapeLength shape) (fun _ -> ofDouble (DiffSharp.Util.Random.Normal()))
+        (values, shape)
+
+    let inline RandomIntegers (maxn: int64) ofDouble (shape:int[]) : (^T[] * int[]) =
+        let values = Array.init (shapeLength shape) (fun _ -> ofDouble (float maxn * DiffSharp.Util.Random.Normal()))
```
gbaydin, line 833: I think this does not implement the correct probabilistic behavior. We can perhaps correct it as:
```fsharp
let values = Array.init (shapeLength shape) (fun _ -> ofDouble (float maxn * DiffSharp.Util.Random.Uniform()))
```
But instead, I think we should be relying on [System.Random.Next](https://docs.microsoft.com/en-us/dotnet/api/system.random.next?view=netcore-3.1#System_Random_Next_System_Int32_System_Int32_) here, adding a little wrapper method called `Integer` to `DiffSharp.Util.Random`.

In any case we need to carefully implement unit tests to ensure the distribution of random integers returned has the correct properties.

dsyme:
OK,this is now green and so is basically ready.  

Only Torch CPU support is enabled - There are now TorchSharp CPU packages available (not yet on nuget.org - we use  [this repository feed](https://donsyme.pkgs.visualstudio.com/TorchSharp/_packaging/packages2/nuget/v3/index.json) - which I think is good enough for our purposes as I can mark packes as retained there)

The remaining TODOs can be done separately:

- reenable Mac support in TorchSharp
- get the TorchSharp CUDA packages pushing reliably 
- add testing for using the CUDA packages, even if only on CPU, just to double check

I haven't done any performance testing, I'll leave that to you @gbaydin

dsyme:
@gbaydin The TorchSharp packages are now available on https://nuget.org - we can pull this when you're ready. The sooner the better from my perspective

dsyme:
@gbaydin I add some stuff to make appveyor pack packages in the CI server and save them with each build.

gbaydin:
Thank you. I agree with handling the overall packaging in a separate PR. I replied in #111 with some detailed comments.

dsyme:
@gbaydin I added `t.gpu()` and `t.cpu()` to the DiffSharp API as part of this, and implemented them for the Torch backend

I've tested `device=CPU` running on the TorchSharp/libtorch CUDA binary set and everything is fine.

I've not tested `device=GPU` and `t.gpu()` as yet since my laptop doesn't have a sufficient GPU.  WIll check if my other machine here does (otherwise I guess I'll start spinning up those Azure instances)

gbaydin:
Ready to merge when latest `dev` branch is integrated.

dsyme:
@gbaydin I've merged dev into this and I think all is good now

src/DiffSharp.Tests/TestDerivatives.fs
```diff
@@ -3620,6 +3620,8 @@ type TestDerivatives () =
     [<Test>]
     member this.TestDerivativeConv3Ds2p2 () =
       for combo in Combos.AllDevicesAndBackends do
+       // 3D dilations not working correctly in LibTorch
```
gbaydin, line 3623: We need to fix this. In the meantime, do we raise an exception to the user when a 3d dilation with libtorch is requested?

gbaydin: I mean, we need to ensure that it is not silently giving incorrect results without any error. Just disabling the test for libtorch is not great because we can forget about fixing the underlying problem in the libtorch backend. 3d convolutions are important to support.

src/DiffSharp.Tests/TestTensor.fs
```diff
@@ -4376,23 +4608,68 @@ type TestTensor () =
...
+            // Dilate 4D 1; 1; 1; 2
+            let tin7 = combo.tensor([[[[1.;2.]; [3.;4.]];[[5.;6.]; [7.;8.]]];[[[1.;2.]; [3.;4.]];[[5.;6.]; [7.;8.]]]])
+            let t7 = tin7.dilate([|1; 1; 1; 2|])
+            let t7Correct = combo.tensor([[[[1.;0.;2.];[3.;0.;4.]]; [[5.;0.;6.];[7.;0.;8.]]]; [[[1.;0.;2.];[3.;0.;4.]]; [[5.;0.;6.];[7.;0.;8.]]]])
+
+            Assert.AreEqual(t7Correct, t7)
+            Assert.AreEqual(combo.dtype, t7.dtype)
+
+            // 3D and 4D dilations not working correctly in LibTorch except when d0, d1 dilations are 1
+            if combo.backend <> Backend.Torch then
```
gbaydin, line 4655: I think disabling tests for a specific backend is not a good idea. We need to fix the backend.

src/DiffSharp.Tests/TestDerivatives.fs
dsyme: Yes, it needs to be fixed - the problem is the Dilate operation as I mentioned.  I think we can do it in a separate PR, or else maybe you could do it in terms of Scatter?

It does raise an error, see here

https://github.com/DiffSharp/DiffSharp/pull/106/files#diff-dca951cab2ee0dc4d6b2ad884ac2e1d8R276

src/DiffSharp.Tests/TestTensor.fs
dsyme: Yup agreed

src/DiffSharp.Tests/TestDerivatives.fs
gbaydin: Ok sounds good. I can look at implementing that in libtorch backend with scatter, while also adding scatter to DiffSharp. I'm opening an issue to remember this. #116

dsyme: Cool, I'll be very glad when that Dilate code goes :-)

dsyme:
I bumped the version of `libtorch-*` and `TorchSharp` to the ones available on nuget.org

gbaydin:
@dsyme I'm merging this now. 

The problem with libtorch dilations have been fixed (with scatter) and all tests are enabled and passing for this backend.

I also added the necessary changes in Model and Data parts to handle different dtypes, devices, backends.

Note that I encountered some excessive memory use when trying out things with this backend, which I concluded are not due to the libtorch backend itself (the issue seems to also affect the reference backend). I would like to explore that in a separate subsequent PR after this. The issue may be linux-specific (e.g., https://github.com/dotnet/runtime/issues/7740, https://github.com/dotnet/runtime/issues/13702)

PR #110, opened 2020-05-18
Optimizers

gbaydin:
This is for keeping track of the generic optimizer APIs


gbaydin:
I think the optimization and model apis are fine for the time being. Merging this now to move on to other things regarding machine learning usage.

Issue #111, opened 2020-05-23
proposal for user choice of packages


dsyme:
Follow on from #106 

After discussing with @gbaydin today my proposal is that we remove the DiffSharp pacakge and have these three "top-level-entry-point" packages:

* package `DiffSharp-torch-cpu`
      (references DiffSharp.Core, DiffSharp.Backends.Torch, libtorch-cpu)

* package `DiffSharp-torch-cuda-10.2-win-x64`
       (references DiffSharp.Core, DiffSharp.Backends.Torch, libtorch-cuda-10.2 for windows)

* package `DiffSharp-torch-cuda-10.2-linux-x64`
      (references DiffSharp.Core, DiffSharp.Backends.Torch, libtorch-cuda-10.2 for linux)

If we have other backends that are high quality and fully feature-complete and tested we can add those too.

With this, the user on a Jupyter notebook simply references one of these and a version number, e.g.

    #r "nuget: DiffSharp-torch-cpu, 1.0.3"

That's a nice and simple choice isn't it? 

Note we really have to split windows/linux Cuda packages here because the downloads are v. large.

Note I haven't listed an accumulation package for the reference backend. I think people wanting to use the reference backend are likely to be building new DiffSharp library components (e.g. derived algorithms) or extensions (new backends etc.) and should reference the individual packages they need.  For example they reference `DiffSharp.Core` in their library project and `DIffSharp.Backend.Reference` in their test project (and a few more packages if they want to test on Torch).  That actually gives a really nice, light experience for creating a library project that adds some stuff to DiffSharp - no need to download anything Torch related if you're happy to trust the reference backend for testing.

Aside : With this, the user gets whatever is the latest version of Torch the `TorchSharp` we depend on happens to work on.  In some far future I could see people wanting to vary the Torch between two or three recent versions, so you can imagine this

* package `DiffSharp-torch-1.5-cpu`
* package `DiffSharp-torch-1.6-cpu`
* package `DiffSharp-torch-1.5-cuda-10.2-linux-x64`
* package `DiffSharp-torch-1.6-cuda-10.3-linux-x64`

etc.  It's an interesting question which of these the user is most likely to want to vary on.  (and will they want to use nightly or modified or patched versions of Torch? That would be hard...)

gbaydin:
Hi @dsyme I have given this more careful thought after our discussion, and the way you summarized things clearly here helped me understand the options better.

Just to list this here for easier reference later, another alternative we can do is:
[...]

PR #117, opened 2020-05-24
Option A of #111

dsyme:
This implements 4 bundling packages as described in https://github.com/DiffSharp/DiffSharp/issues/111#issuecomment-633106778. Also adds `dsharp.config(...)` as discussed there.

At the moment you have to set

    dsharp.config(Backend.Torch)

manually even if the Torch backend is the only one referenced.   It's somewhat awkward and error-prone to remove that as it would require determining that there's only one backend through some fairly error prone techniques (e.g. scanning available assemblies etc.).  All


dsyme:
> I've marked this as "work in progress" because there is one problem...

I've fixed this problem, the download of cuda is no longer required to create the relevant packages.

dsyme:
@gbaydin this is ready when you're happy with it.

PR #119, opened 2020-06-02
small fixes for GPU torch

dsyme:
Various fixes from this thread: https://github.com/xamarin/TorchSharp/issues/143

@gbaydin I think these can go straight in as long as they are green for CPU


dsyme:
@gbaydin I will merge these directly as they are simple fixes for GPU Torch. Please review post-hoc?

gbaydin:
Looks fine for the time being, although we need to support devices like `cuda:0`, `cuda:1`, etc. in the near future

PR #120, opened 2020-06-02
move scalars to gpu

dsyme:
Part of fixing a bunch of issues with Torch GPU backend in https://github.com/xamarin/TorchSharp/issues/143

In this case scalars were not being converted to GPU tensors.


dsyme:
Merging as it is a simple fix for Torch GPU based on systematic testing in https://github.com/xamarin/TorchSharp/issues/143 (which we will be able to enable once we can run CI on GPU)

gbaydin:
Looks fine

PR #121, opened 2020-06-02
When serializing a tensor save the right device

dsyme:
When serializing a tensor we save the device it is associated with, and resurrect it back to that device.

One of the last bugs detected by this test pass for libtorch on GPU https://github.com/xamarin/TorchSharp/issues/143

### DIFF
--- a/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
+++ b/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
@@ -750,9 +750,9 @@ type TorchRawTensor(tt: TorchTensor, shape: int[], dtype, device) =
             // Torch Tensors must be CPU before they can be saved
             let tCpu = t.MoveTo(Device.CPU) :?> TorchRawTensor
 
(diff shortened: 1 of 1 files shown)


pkese:
@dsyme I also tried to fix this albeit with slightly different approach:

https://github.com/pkese/DiffSharp/commit/cde4b95716ad434080b6cfd5398968f95177bb2d

I set the backend and device to curently configured model/backend after model is loaded.

This makes it more natural to save model made on CPU and load it somewhere else where GPU is available and configured (without having to be aware whether model was saved on computer with CPU, GPU or Reference implementation).

On the other hand it does make sense to save `dtype` into model file.

BTW, all tests now pass 🎉

dsyme:
> This makes it more natural to save model made on CPU and load it somewhere else where GPU is available and configured (without having to be aware whether model was saved on computer with CPU, GPU or Reference implementation).

I will add a separate issue about whether the serialization format for a tensor should record device, dtype, backend.

dsyme:
> BTW, all tests now pass 🎉

Fantastic!

PR #127, opened 2020-06-04
update torchsharp version

dsyme:
1. Updates the `libtorch` nuget
2. Updates the `TorchSharp` nuget
3. Removes the reference to two incompatible runtimes in `tests\Test`

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,8 +2,8 @@
 
   <PropertyGroup>
 
-     <LibTorchVersion>1.5.0</LibTorchVersion>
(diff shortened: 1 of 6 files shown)


dsyme:
@gbaydin I will merge this as it is better to rely on these packages rather than older ones.

The linux cuda packages should now work more reliably (I hope) - or if there is an error in reconstituting them then there should be a decent error report.  In that situation you might need to 

1. try again with the build or restore
2. nuke `~/.nuget/packages/libtorch*` and try again.

gbaydin:
Sounds good, thank you!

Issue #131, opened 2020-06-06
Incorrect reverse operations for multiplication with zero-dimension tensors


VisualMelon:
Products of scalars break reverse mode AD as either side of a multiplication. Example:

module DiffSharpTest.Program

```
open DiffSharp
open DiffSharp.Model

type ConstModel(size: int) =
    inherit Model()
    let b = Parameter <| dsharp.rand [size; size]
    do base.add([b], ["b"])
    override c.forward(value) =
        b.value
    member public this.B with get() = b

[<EntryPoint>]
let main argv =

    dsharp.config(backend=Backend.Reference)
    dsharp.seed(1)

    let model = ConstModel(10)
    model.reverseDiff()
    let loss = model.B.value.sum() * dsharp.tensor [1.0] // sum gives us a scalar: without it, everything works as you would expect
    loss.reverse()
    printfn "Grad: %A" model.B.value.derivative
    0
```

The gradients on `model.B` should be a matrix of ones, but instead it produces a scalar zero. This seems to be resolved by patching [`(*)`](https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Core/Tensor.fs#L624-L643) as in https://github.com/VisualMelon/DiffSharp/commit/a3c599510a53d11d329e5dac85db6aa236384349, but I don't trust that I'm using the right 'overloads' still.

Other operations also look suspect in the `dim=0` case, e.g. [`(/)`](https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Core/Tensor.fs#L680), but I'm not convinced I understand the T/T0 distinction properly yet, and come to a different conclusion every time I look at it, so I'm posting this before I end up in any more of a tangle. I'd be happy to do sort out a PR once I'm convinced I know what's going on.

These presumably will correspond to the currently missing tests, e.g. https://github.com/DiffSharp/DiffSharp/blob/dev/tests/DiffSharp.Tests/TestDerivatives.fs#L231-L234, which I'll look into writing

gbaydin:
Thank you!

PR #132, opened 2020-06-07
Additional tests for addition and multiplication

VisualMelon:
Adds additional tests for addition, multiplication, subtraction, division, and exponentiation. 

Fixes 2 issues, where reverse mode differentiation between scalars and constants resulted in incorrect derivatives (see #131)

I couldn't find any contributor guidelines: shout if you want this broken up into separate PRs etc.

### DIFF
--- a/src/DiffSharp.Core/Tensor.fs
+++ b/src/DiffSharp.Core/Tensor.fs
@@ -628,7 +628,7 @@ type Tensor =
             let dfTensorFwdTC(cp:Tensor,ap:Tensor,ad:Tensor) = ad * b
             let dfTensorFwdCT(cp:Tensor,bp:Tensor,bd:Tensor) = a * bd
             let dfTensorRevTT(a,b) = MulTT0(b,a)
-            let dfTensorRevTC(a,b) = MulTConstT0(a,b)
+            let dfTensorRevTC(a,b) = MulTConstT0(b,a)
             let dfTensorRevCT(a,b) = MulTT0Const(b,a)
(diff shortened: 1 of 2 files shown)


gbaydin:
Hi @VisualMelon, thank you for noticing this. Your fixes in [Tensor.fs](https://github.com/DiffSharp/DiffSharp/pull/132/commits/04a9896535d45f647f6e2f2cb39d4797ba7bd959) look correct to me. I independently inspected the code and came up with the same conclusion.

Could you please tell me how you derived the correct values for the test cases in [TestDerivatives.fs](https://github.com/DiffSharp/DiffSharp/pull/132/files#diff-866f95b694d1b5a956687f0eb535c95b)? I will need to check them analytically and also with [difftorch](https://github.com/gbaydin/difftorch) and perhaps modify/add some to this PR before merging, if this is fine with you.

VisualMelon:
@gbaydin The add, mul, and sub stuff was easy to compute as went. For div I had to do the derivations on paper (because apparently I've not done this for a while) and took somewhat longer to enter into powershell (which I used to perform the arithmetic), so for `pow` I wrote a script to compute the derivates for a given combination of x,dx,y,dy,dz and sanity checked them as I entered them (manually again). All the numbers were copied from the calculations in powershell, not computed with the library. The `Const` variants then just require deleting reverse mode stuff and recomputing the forward mode `dz`. The `swap` stuff saved duplicating code for the symmetric ops; I hope it isn't too objectionable.

gbaydin:
> I appreciate that it's important to verify these: if I can help with that at all, let me know. Any little bug in here could create untold misery down the line (I was lucky that the error in mul broke my code rather than just producing incorrect values).

Exactly! Thank you, I'm merging this now and I will verify the tests and make changes if needed.

PR #133, opened 2020-06-07
Debug/memory

gbaydin:
@dsyme I think I identified the source of the excessive memory usage we witnessed in some code after merging the libtorch backend to dev.

The problem seems to be related with not disposing `TorchTensor` objects and freeing the associated unmanaged memory. The `MemoryTest` project in this PR provides an isolated minimal example. The problem is only with the libtorch backend and is not present in the Reference backend.

The core of the problem is
```fsharp
    dsharp.config(backend=Backend.Reference) // Uses around 23 MB
    //dsharp.config(backend=Backend.Torch) // Uses around 4GB

    for i in 0..1024 do
        let _ = dsharp.zeros([1024; 1024])
        printfn "%A" i
```

When run with the libtorch backend, the memory usage of this example is monotonically rising and reaches around 4GB:

When run with the Reference backend, it is around 20 MB (and the managed heap seems to hold around 4MB which is consistent with what one would expect from holding a single instance of a 1024x1024 single precision tensor). I can see the memory use fluctuate around this value as Tensors get created and discarded.

This behavior is the same on Windows and Linux (I tested both).

The problem remains if I remove DiffSharp from the code and just use TorchSharp. The following uses around 4GB memory.
```fsharp
    for i in 0..1024 do
        let x = TorchSharp.Tensor.FloatTensor.Zeros([|1024L;1024L|])
        //x.Dispose()
        printfn "%A" i
        System.GC.Collect()
```
The problem disappears when I manually do `x.Dispose()`.

### DIFF
--- a/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
+++ b/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
@@ -50,6 +50,7 @@ module internal Utils =
 /// This is the base class for all RawTensorXyz tuypes.
 /// All type-independent operations are implemented directly on this class. 
 type TorchRawTensor(tt: TorchTensor, shape: int[], dtype, device) =
+
     inherit RawTensor(shape, dtype, device, Backend.Torch)
 
     do 
@@ -58,12 +59,6 @@ type TorchRawTensor(tt: TorchTensor, shape: int[], dtype, device) =
 
        if toTorchShape shape <> tt.Shape then 
            failwithf "mismatched Torch tensor shape, expected %A, got %A" (toTorchShape shape) tt.Shape
-
-    override _.Finalize() =
-        try
-            tt.Dispose()
-        finally
-            base.Finalize()
 
     member t.MakeLike(tt, ?shape, ?dtype, ?device) : RawTensor =
         upcast TorchRawTensor(tt, defaultArg shape t.Shape, defaultArg dtype t.Dtype, defaultArg device t.Device)
@@ -79,7 +74,7 @@ type TorchRawTensor(tt: TorchTensor, shape: int[], dtype, device) =
             let stop = fullBounds.[i,1] + 1
 
             let len = stop - start
-            let idxs = LongTensor.Arange(int64 start, int64 stop, 1L, device=toTorchDevice t.Device)
+            use idxs = LongTensor.Arange(int64 start, int64 stop, 1L, device=toTorchDevice t.Device)
             res <- res.IndexSelect(int64 dim, idxs)  // yield len // if len=1 then squeeze this dimension
             if fullBounds.[i, 2] = 1 && len = 1 then 
                 res <- res.Squeeze(int64 dim)  // yield len // if len=1 then squeeze this dimension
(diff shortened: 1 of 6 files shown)


gbaydin:
This discussion seems relevant: https://stackoverflow.com/questions/30517301/will-net-call-dispose-for-me-in-this-case

gbaydin:
Well good news, just implementing
```fsharp
override t.Finalize() =
    tt.Dispose()
```
in `TorchRawTensor` seems to solve the problem. I'm inspecting it further to make sure.

gbaydin:
My latest commit seems to fix the issue. Maybe the problem is ultimately because https://github.com/xamarin/TorchSharp/blob/1e180589226bdd7a0f823c2eab97c02ec4395c18/src/TorchSharp/Tensor/TorchTensor.cs does not implement a destructor as described [here](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/destructors).

gbaydin:
Actually it seems a bit more complicated than that. I'm getting "double free" errors because of trying to free the same resources twice. I think this may have to do something with memory sharing `TorchTensor`'s in the `TorchRawTensor` implementation.

dsyme:
Good work, I'll take a look at this.  The basic fix should be ok but we should find the cause of the double free

dsyme:
I added some corresponding tests to TorchSharp to make sure we are disposing at that level. 
We can add more tests there as needed

https://github.com/xamarin/TorchSharp/pull/153

dsyme:
@gbaydin I think this will effectively fix things: https://github.com/xamarin/TorchSharp/pull/154

gbaydin:
Sounds great, looking forward to that. Once we update the code in this branch to use the new package we should check whether the tests pass. Currently they are failing because of the "double free" error. 

I think we also need to remove the `TorchRawTensor.Finalize()` code from DiffSharp now.

gbaydin:
I will also add a memory usage unit test to DiffSharp to ensure we get notified if some memory problem appears in the future.

dsyme:
This should now be fixed byt he latest update in the PR.  Once green I will merge it

dsyme:
> I will also add a memory usage unit test to DiffSharp to ensure we get notified if some memory problem appears in the future.

This is done by my update

src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
```diff
@@ -73,7 +74,7 @@ type TorchRawTensor(tt: TorchTensor, shape: int[], dtype, device) =
             let stop = fullBounds.[i,1] + 1
 
             let len = stop - start
-            let idxs = LongTensor.Arange(int64 start, int64 stop, 1L, device=toTorchDevice t.Device)
+            use idxs = LongTensor.Arange(int64 start, int64 stop, 1L, device=toTorchDevice t.Device)
```
dsyme, line 77: This line is not essential.  But it indicates how we can get deterministic disposal of tensors when we need to (the `Arange` tensor and its backing unmanaged resources are disposed as soon as `idxs` is no longer in scope).  This is likely to become important for performance as time goes on.

gbaydin: Thank you, this is a good example.

gbaydin:
The part I don't undertand about the test case

```fsharp
    [<Test>]
    member _.TestTensorZerosDisposal () =
        for i in 0..1024 do
            let _ = dsharp.zeros([1024; 1024])
            printfn "%A" i
            //System.GC.Collect()
```

is that it's not really checking for a condition, and it can pass even when we use 4GB of memory during testing. I was thinking about actually monitoring the memory usage and having an assertion that expects the memory used to be below some threshold.

PR #138, opened 2020-06-12
trial docs using FSharp.Formatting and .NET Core

dsyme:
I trialled creating docs using FSharp.Formatting 4.1 and .NET Core, it works just fine

1. I took the docs content from master
2. I created a docs\BuildDocs application referencing FSharp.Formatting (instead of the docs\BuildDocs.fsx).  This is an application which I build and run manually to generate the docs.
3. I edited the scripting content of the docs to adjust for the new API, mainly by using a [`helpers.fsx`](https://github.com/DiffSharp/DiffSharp/pull/138/files#diff-6c084362c3e639880040ca41086e7b95)

To show the script docs are being generated using the new API, note the use of `Tensor` below:

image

and the API reference docs

image


dsyme:
@gbaydin This is mainly a trial to check that it is possible and the results are at least as good as today.

The content obviously needs complete revision but I think this may be the best way to go.   We would need to 

1. Freshen the visual look in the templare
2. Update all content
3. Also generate .NET jupyter notebooks from the scripts

There are a lot of things I noticed along the way about the API and content, I'll raise those in a separate issue though I'm sure you'll notice them as well.

dsyme:
@gbaydin I think I'd recommend we go with this as our documentation solution.  It's familiar enough to us both and includes F# literate scripting.

FSharp.Formatting has seen some love recently and our use of it is simple enough (and doesn't require any use of Fake or Paket)

dsyme:
> @gbaydin I think I'd recommend we go with this as our documentation solution. It's familiar enough to us both and includes F# literate scripting.

1. That doesn't mean you have to accept any of the docs as I've ported them here - that was really just a trial to see how it went. I could remove all of those (apart from a `index.fsx` I guess) from this PR and just leave the CreateDocs app and the reference docs.  Or we could start with what we have here and iterate.

2. We can later use a CreateDocs.fsx script once some bugs are fixed with `dotnet fsi`.

3. We should move to XPlot for all plotting in the samples.

dsyme:
@gbaydin I'm doing a lot of basic work on FSharp.Formatting, which is used by pretty much all formatting engines to format docs all F# code, to modernise and cleanup the generated docs, e.g. https://github.com/fsprojects/FSharp.Formatting/pull/527

Am also looking to include searchable docs based on aligning with related improvements by @Krzysztof-Cieslak in Waypoint

dsyme:
Progress

* Revamped FSharp.Formatting so it does really simple content-driven doc generation

* Source is docs/... , specifically *.md and *.fsx and templates for HTML, ipynb and simplified fsx outputs

* Output docs go in `gh-pages`

* Binder buttons opening notebooks in mybinder available for every generated Script and Markdown:

image

dsyme:
@gbaydin This is pretty much ready as a first step in documentation generation

We can walk through it on Friday.  

* The `docs` directory is pretty much self explanatory.  It's totally content driven.  

* The process for building them manually is described in docs/README.md.  It's quick turnaround.

* The doc generation is not yet part of the main build, it's done manually and pushed to the URL I mentioned to you by email.   

* Currently it relies on building `FSharp.Formatting` master branch in a parallel directoy. I'll get the FSharp.Formatting packages pushed tomorrow.

PR #145, opened 2020-06-23
float16/bloat16 support

dsyme:
This is preliminary support for `float16`, `bfloat16` tensors.  

Note that values are read/written in/out of the tensors as `float32`.  I'd expect us to have to take a similar approach for quantized integers as well when we eventually support those.


gbaydin:
Some relevant information about the numeric types relevant/useful for ML workloads currently

gbaydin:
Some up-to-date information here in this update about Nvidia Ampere GPUs about the numeric types relevant/useful for current workloads: 

"The new Ampere GPU architecture for our A2 instances features several innovations that are immediately beneficial to many ML and HPC workloads. A100’s new Tensor Float 32 (TF32) format provides 10x speed improvement compared to FP32 performance of the previous generation Volta V100. The A100 also has enhanced 16-bit math capabilities supporting both FP16 and bfloat16 (BF16) at double the rate of TF32. INT8, INT4 and INT1 tensor operations are also supported now making A100 an equally excellent option for inference workloads. Also, the A100’s new Sparse Tensor Core instructions allow skipping the compute on entries with zero values, resulting in a doubling of the Tensor Core compute throughput of int8, FP16, BF16 and TF32. Lastly, the multi-instance group (mig) feature allows each GPU to be partitioned into as many as seven GPU instances, fully isolated from a performance and fault isolation perspective. All together, each A100 will have a lot more performance, increased memory, very flexible precision support, and increased process isolation for running multiple workloads on a single GPU. "
https://cloud.google.com/blog/products/compute/announcing-google-cloud-a2-vm-family-based-on-nvidia-a100-gpu

dsyme:
@gbaydin I've updated bfloat16/float16 support in TorchSharp, and this PR now activates that support in DiffSharp.

That's the good news.  However, there are many cases where LibTorch is missing functionality for either bfloat16/float16 or both, so there are places where this support will hit missing functionality.  

You can see these in the test suite where `Combos.*Except16s` is used in about 50 places.

Most of these represent functionality that can be activated incrementally, either by

1. We use float32 instead then convert to float16/bfloat16 where necessary, this is already done in a couple of places; OR 

2. We adjust the test case (e.g. to avoid random number generation, which is not directly supported for float16)

The refrence backend just uses float32 as replacement for float16/bfloat16

I think this is good enough to integrate as a first step towards more uniform support - what works is pinned down and tested, and we can iterate from here.

tests/DiffSharp.Tests/TestDistributions.fs
```diff
@@ -195,7 +195,7 @@ type TestDistributions () =
                 let eventShapeCorrect = [||]
                 let probs = d.probs
                 let logits = d.logits
-                let samples = d.sample(10*numEmpiricalSamples)
```
dsyme, line 198: adjusted to make this test run faster, it was one of the longest running tests

tests/DiffSharp.Tests/TestModel.fs
```diff
@@ -210,20 +210,18 @@ type TestModel () =
 
     [<Test>]
     member _.TestModelMove () =
-        let confBackup = dsharp.config()
-        for combo1 in Combos.FloatingPoint do
-            dsharp.config(combo1.dtype, combo1.device, combo1.backend)
+        for combo1 in Combos.FloatingPointExcept16s do
```
dsyme, line 213: adjusted to make this test run faster, it was one of the longest running tests

dsyme:
@gbaydin, I have marked `Dtype.BFloat16` and `Dtype.Float16` as experimental, so people know support is limited.

Merging to get this baseline support for bfloat16/float16 into the codebase.

PR #146, opened 2020-06-25
allow all torch devices + device indexes

dsyme:
This does:

* [x] updates TorchSharp
* [x] You can now specify a device type + device index with 
* [x] adds `dsharp.devices()` to get the supported devices, and `dsharp.devices(backend)` to get them for a particular backend

I will actually run tests on CUDA tomorrow on my local machine.  It's all set up to do it just need to double check everything.

Examples of intended use:
```fsharp
let t = dsharp.tensor(3)

// default devices 
t.move(Device.CPU)   //= Device(DeviceType.CPU, 0)
t.move(Device.GPU)   //= Device(DeviceType.GPU, 0)

// indexed devices 
t.move(Device(DeviceType.CUDA, 4))
t.move(Device(DeviceType.CPU, 4))
t.move(Device(DeviceType.CPU, 4))

// query for available devices 
dsharp.devices() 
dsharp.devices(DeviceType.CUDA) 
dsharp.devices(backend=Backend.Torch) 

dsharp.isDeviceTypeSupported(DeviceType.CUDA)
dsharp.isDeviceTypeSupported(DeviceType.XLA) 
dsharp.isDeviceTypeSupported(DeviceType.CUDA, backend=Backend.Torch)

// for convenience
dsharp.isCudaSupported()     // = dsharp.isDeviceTypeSupported(DeviceType.CUDA)

```
Note there is no support for specifying devices as strings:
```
t.move(Device("cuda:0")) // not allowed at the moment
```


dsyme:
Note I've just copied the list of all possible devices from Torch here, I think that seems an OK approach for now.  Each backend reports which specific devices are supported

https://github.com/DiffSharp/DiffSharp/pull/146/files#diff-7fa361f38182684676724866a74fc9e2R5

dsyme:
@gbaydin weirdly Travis is not showing in the set of passing checks, though it is passing on Travis here: https://travis-ci.org/github/DiffSharp/DiffSharp

I'm not sure wha's changed, we might need to re-establish the hooks

gbaydin:
Ok I will have a look. I haven't changed any settings. I suspect it might have to do with the recent GitHub changes (which I thought were visual changes only).

dsyme:
@gbaydin This is now ready.  

- All tests now pass for CUDA 10.2 on Windows

- Full range of pytorch devices supported, plus device indexes

src/DiffSharp.Core/Tensor.fs
```diff
@@ -42,7 +42,13 @@ type Tensor =
         | TensorF(_) -> failwith "Cannot cast TensorF - do not cast during differentiation"
         | TensorR(_) -> failwith "Cannot cast TensorR - do not cast during differentiation"
 
-    member t.move(backend) =
+    member t.move(backend: Backend) =
+        // If a backend move is needed then first move to the CPU
+        let t =
```
dsyme, line 47: @gbaydin Just to explain one things here

Initially in this PR I adjusted the backend to mimic supporting both GPU and CPU.  This would allow us to better pin down the expected behaviour for things like moving between devices and backends.  However it doubles tests times so I took it out

However that did catch a case where the DataLoader move test was moving data from "Reference GPU" to "Torch CPU".  

In cases of a backend move I think we should always move to CPU first, and then move backends, and then to the target device in the other backend.

dsyme:
OK, this is now ready

dsyme:
@gbaydin Can we pull this in now? Thanks

dsyme:
@gbaydin If it's ok I'm going to pull this now, can you do a post-hoc review on it and I'll fix up anything you spot?

dsyme:
@gbaydin Merging this now

gbaydin:
@dsyme thank you, I'm reviewing this now.

gbaydin:
@dsyme everything looks good to me. 

One minor thing we should consider adding is to support constructing Device objects with strings, as you noted above.

```fsharp
t.move(Device("cuda:0")) // not allowed at the moment
```

Issue #152, opened 2020-07-13
Default printing of tensors in notebooks causes exception


dsyme:
The default printing of scalar tensors in F# .NET Interactive notebooks causes an exception:

1. create an F# notebook at https://mybinder.org/v2/gh/dotnet/interactive/master?urlpath=lab

2. Run this

```
#i "nuget: https://ci.appveyor.com/nuget/diffsharp"
#r "nuget: DiffSharp-lite,0.9.5-preview-2006281542" // update version as needed, see a recent build in https://ci.appveyor.com/project/dsyme/diffsharp/history

open DiffSharp

dsharp.tensor 3.0
```

Expected: nice printing of tensor

Actual: exception "Cannot enumerate Tensor with dim "

Oddly also applies to tensors of any dimension.

```
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1639.Invoke(String message) in F:\workspace\_work\1\s\src\fsharp\FSharp.Core\printf.fs:line 1639
   at DiffSharp.Tensor.System-Collections-Generic-IEnumerable`1-GetEnumerator()
   at System.Linq.Enumerable.SelectIterator[TSource,TResult](IEnumerable`1 source, Func`3 selector)+MoveNext()
   at Microsoft.DotNet.Interactive.Formatting.EnumerableExtensions.TakeAndCountRemaining[T](IEnumerable`1 source, Int32 count) in F:\workspace\_work\1\s\src\Microsoft.DotNet.Interactive.Formatting\EnumerableExtensions.cs:line 19
   at Microsoft.DotNet.Interactive.Formatting.HtmlFormatter`1.<>c__DisplayClass7_0.<CreateForSequence>g__BuildTable|3(T source, TextWriter writer) in F:\workspace\_work\1\s\src\Microsoft.DotNet.Interactive.Formatting\HtmlFormatter{T}.cs:line 112
```

dsyme:
@gbaydin The problem stems from this:

https://github.com/DiffSharp/DiffSharp/commit/87cc85ca690344272046df4ed38196f08a28e50a

and the decision to have `Tensor` implement `IEnumerable<Tensor>`.  

I know what you're trying to do here but I believe this isn't really going to fly. There's just a lot of .NET tooling that takes values and attempts to do some kind of display on them.  It inevitably means that using "enumerate a collection" logic which is all over .NET will fail when it eaches the base case of scalars.

[...]

PR #153, opened 2020-07-14
Rollback Tensor IEnumerable

dsyme:
#152 documents how making Tensor IEnumerable brings problems with any .NET display code that assumes it can enumerate IEnumerable things for display purposes, perhaps recursively 

For now I think we should roll back the addition if the IEnumerable and reconsider later


dsyme:
@gbaydin Are you ok with this?  We really need it or printing in notebooks will be very problematic

gbaydin:
@dsyme I'm looking at this and I'm going to reply in #152. Let's continue the conversation there.

gbaydin:
Merging this following the discussion at #152

PR #154, opened 2020-08-06
doc tooling update

dsyme:
I've been knocking the F# doc tooling into shape at http://fsprojects.github.io/FSharp.Formatting/ and proofing it on the docs for FSharp.Core https://fsharp.github.io/fsharp-core-docs/

We're now able to pick up those improvements in this repo and start applying them to simplify the DiffSharp docs


dsyme:
@gbaydin (I actually didn't mean to merge this, though it won't be a problem)

PR #157, opened 2020-08-26
Use dsharp name by default for operations like dsharp.tensor, dsharp.rand, dsharp.stack, etc.

gbaydin:
This is making the `dsharp` name the default and only name for things like `dsharp.tensor [1,2,3]`. Previously one could also do `DiffSharp.tensor [1,2,3]`. The `dsharp` name is lightweight and fits better with the rest of the (PyTorch-inspired) naming in place. I've been using it in the code throughout and it feels very natural.

@dsyme could you please confirm you're ok with this? I remember we had a conversation about it and decided to go with this change. I think the main motivation was an issue with dotnet interactive. Could you please remind me if you remember? :)

### DIFF
--- a/src/DiffSharp.Core/DiffSharp.fs
+++ b/src/DiffSharp.Core/DiffSharp.fs
@@ -5,7 +5,7 @@ open DiffSharp.Util
 
 /// Tensor operations
 
-type DiffSharp =
+type dsharp =
     static member tensor(value:obj, ?dtype:Dtype, ?device:Device, ?backend:Backend) = Tensor.create(value=value, ?dtype=dtype, ?device=device, ?backend=backend)
     static member seed(?seed:int) = BackendStatics.Seed(?seed=seed)
     static member isTensor(value:obj) = value :? Tensor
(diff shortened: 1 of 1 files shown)


gbaydin:
Merging this now. Please let me know if there are any objections.

src/DiffSharp.Core/DiffSharp.fs
```diff
@@ -287,212 +287,209 @@ type DiffSharp with
...
-    static member numfgh f x = DiffSharp.numfgradhessian f x
-    static member numfjvp f x v = DiffSharp.numfjacobianv f x v
-    static member numfj f x = DiffSharp.numfjacobian f x    
-
-
-type dsharp = DiffSharp
\ No newline at end of file
+type dsharp with
+    static member numgvp f x v = dsharp.numgradv f x v
+    static member numg f x = dsharp.numgrad f x
```
dsyme, line 483: This all feels like we're abbreviation abbreviations, it feels a bit to much to be honest 

Perhaps there should be abbreviations and full naming and no half-abbreviations.

gbaydin: I think these names and api design can be revised as we talked about some time ago. For example, instead of the `num` prefix in these numerical differentiation operations, we can have a separate namespace, etc. I think it's worth handling in a dedicated pull request.

dsyme: That's not really an objection, just a comment. You might want to get this part of the API assessed by a target expert currently using torch? It's just very information-dense and I think people (esp those just doing tensor programming, not writing optimizers) might be overwhelmed by the mass of abbreviations and possibilities here

I mean how important are the abbreviations anyway? There are only ever a couple of calls to these functions per optimizer right? And they are crucial calls. It seems an area where people spell might benefit from full names.

dsyme: (it's fine for now, just keep this in mind)

gbaydin: I think if things are well-documented in DiffSharp docs and tutorials, I think it's fine to have these abbreviated aliases in addition to the "classical" DiffSharp API names like jacobianv, gradhessian, etc. 

I'm not a fan myself, but things like `jvp`, `hvp` etc. are here to stay. This is mostly gaining popularity due to Google projects.

gbaydin: I think we can separate these functional autodiff apis or abbreviated versions to another location other than `dsharp.` so that regular users don't get too exposed to these in intellisense, etc. But personally I think it's also fine if they are exposed. They can discover, read the docs, learn a few things about the capabilities etc.

VisualMelon:
I would note that `dsharp` is out of place in .NET land, and python users are used to having stuff like `type dsharp = DiffSharp.DiffSharp` at the top of every single file/notebook.

dsyme:
> I would note that dsharp is out of place in .NET land, and python users are used to having stuff like type dsharp = DiffSharp.DiffSharp at the top of every single file/notebook.

We know, it's been discussed. The API is not being design to follow .NET conventions, and instead follows largely pythonesque naming. This is similar to most tech in [SciSharp](https://github.com/SciSharp)

DiffSharp is a tensor technology that happens to use F# and the .NET runtime.  It is not a .NET/C# Microsoft.Math.TensorLibrary and its target audience is not C# programmers.  Instead the target ausience is, roughly speaking, are AI researchers who are pretty much happy with Python programming, but who want strong typing and some other very nice things.

PR #158, opened 2020-08-26
Neural network models and workflows

gbaydin:
This is to keep track of the next steps in improving coverage of standard neural networks and workflows. Starting with variational autoencoders (VAEs), moving on to generative adversarial networks (GANs), recurrent neural networks (LSTM, GRU), transformers, etc.


dsyme:
Nice work so far

PR #168, opened 2020-09-07
Add docs to DiffSharp.Util

dsyme:
* Start adding the `///` docs across the API

* `Shape` is moved into `DiffSharp` namespace and add Shape.fs to hold all shape-related logic

* `DiffSharp.Util` becomes a namespace 

Separately I think we should rename

    DiffSharp.Util --> DiffSharp.Utilities
    DiffSharp.Optim --> DiffSharp.Optimizers

and make some of DiffSharp.Utilities internal


dsyme:
@gbaydin I will merge this.  It is just shuffling some code around under `Diffsharp.Util` and splitting some files.

The "patch" code coverage test failed but that's because it is moving code around

Issue #169, opened 2020-09-07
Confusion regarding DiffSharp-cuda-windows bundle


VisualMelon:
I can't get the DiffSharp-cuda-windows bundle to work. Three things that confuse me:
 1.  It depends on `libtorch-cuda-10.2-linux-x64` (rather than `libtorch-cuda-10.2-win-x64`), which naturally doesn't contain the necessary DLLs to run under windows
 2. Any attempt to restore it seems to fail, and I don't understand why; the DiffSharp-cpu bundle restores just fine, and looks mostly the same. The error is "Package 'DiffSharp-cuda-windows' is incompatible with 'all' frameworks in project". This doesn't seem to be transient or machine specific problem, but I don't understand it, so it might just be me.
 3. It targets netcoreapp3.0; I'd have expected it to be netstandard2.1 (per the nuspec); however, I see this was explicitly changed recently, so I assume there is a reason for this.

On a positive note, I have successfully got CUDA/GPU working under windows by manually adding the following dependencies (the newly added CI builds on nuget):
 - DiffSharp.Backends.Torch Version="1.0.0-preview-243323135"
 - DiffSharp.Core Version="1.0.0-preview-243323135"
 - libtorch-cuda-10.2-win-x64 Version="1.5.6"

Incidentally, I'm probably going to end up using DiffSharp for significantly more work in the near future, and I recall something about testing under Windows GPU being a problem: I'd be happy to put some effort in if it's something I can help with (my main machine is Win 10 with an RTX 2060 Super).

dsyme:
Thanks should be fixed by https://github.com/DiffSharp/DiffSharp/pull/170

VisualMelon:
Great, I'll report back once that gest to the CI and I find a minute to test it.

dsyme:
> Incidentally, I'm probably going to end up using DiffSharp for significantly more work in the near future, and I recall something about testing under Windows GPU being a problem: I'd be happy to put some effort in if it's something I can help with (my main machine is Win 10 with an RTX 2060 Super).

We currently do manual testing on GPU from time to time on roughly this kind of machine.  Last time I checked we had a couple of tiny errors in test, see https://github.com/DiffSharp/DiffSharp/issues/147

Anything you can do to either automate or manually run tests (and fix them up, report on status etc.) would be very much appreciated!!

dsyme:
@VisualMelon We'd also love any GPU testing on Linux you happen to be able to muster

[...]

PR #171, opened 2020-09-08
fix packages

dsyme:
Another stab at fixing https://github.com/DiffSharp/DiffSharp/issues/169


.github/workflows/publish.yml
```diff
@@ -33,6 +33,10 @@ jobs:
       uses: codecov/codecov-action@v1
     - name: Pack
       run: dotnet pack --configuration Release --verbosity normal
+    - name: Pack (DiffSharp-cuda-linux)
+      run: dotnet pack --configuration Release --verbosity normal bundles\DiffSharp-cuda-linux
```
VisualMelon, line 37: Needs to be `/` (same below): causing the packaging to fail: https://github.com/DiffSharp/DiffSharp/runs/1086050946?check_suite_focus=true#step:11:10

dsyme: Thanks!  fixed and CI is trying again :-)

PR #172, opened 2020-09-08
some micro perf testing and tweaks

### LINKED ISSUE #124: Backends - should not RawTensor capture device etc.
Current `RawTensor` and `TorchRawTensor` are capturing a number of values in addition to the actual data (that is, these values become fields of these objects)

* RawTensor captures all of   `shape`, `dtype`, `device`, `backend`

* I think TorchRawTensor captures `tt: TorchTensor`, `shape: int[]`, `dtype`, `device` independently causing some duplication

If we wish we can get all this information from the TorchTensor itself, in which case all of these should be accessed via abstract properties implemented in terms of `tt` 

Part of the question is how much we care about the size of `Tensor`, `RawTensor` and `TorchRawTensor`.  We should likely work to reduce these to the absolute minimum.  

Currently for each DiffSharp Tensor value we have the following:  
* `Tensor  - a discriminated union which for Primals not undergoing differentiation is a single allocation around a...
*  `TorchRawTensor + RawTensor` which wraps the ...
* TorchSharp `TorchTensor` which is a zero-cost struct wrapping that handle for a...
* `THSTensor`, a C++ shared pointer that wraps the actual tensor value

Another factor is whether `Shape` is `int64[]` or `int32[]` - TorchTensor prefers the former and we would repeatedly reallocate it.

dsyme:
I looked into some perf characteristics for #124

1. Fixes #124  - RawTensor no longer stores shape, device etc. but gets it from the implementation

2. Removes "device" field from `TorchRawTensor` and gets it from the underlying TorchTensor instead

3. Adds a Benchmark.NET test program using CUDA


dsyme:
I took some measurements of repeated "tensor + tensor" and approximately fitted them to a cost model based on 

* N = overall number of primitive additions
* n = number of elements in the tensor
* t = c1 * N + c2 / (N/n)
* c1 = u/s per primitive addition
* c2 = u/s per overall tensor addition operation (not including creating tensors nor getting data to the GPU)

Very approximate figures

* Torch CPU, best c1 = 0.0010, c2 = approx 8.0

* Torch GPU, best c1 = 0.000056, c2 = approx 75.0

* Reference CPU, best c1 = 0.0025, c3 = approx 1.0 

These are pretty much as you'd expect 

* setting up operations on the GPU is expensive then very fast

* the reference backend has low overhead to reach down into the actual tensor data but is significantly slower on actual floating point performance 

* overall we don't expect DiffSharp to be fast on tiny tensors.  Indeed for this particular operation it's not until you have about tensors of about size 10,000 that the Torch CPU or GPU backends become faster than the current reference backend (note however that the addition operation is implemented in a tight loop - many other operations in the reference backend are far less efficient)

I'll put this in a doc to revisit it later

dsyme:
@gbaydin I'll merge this, it's all fairly straight-forward

dsyme:
@gbaydin Some thigns to note

- using Benchmark.NET for performance benchmarking

- CPU has device index -1 in Torch so we'll do the same thing in DiffSharp

VisualMelon:
The test `TestTensorSaveLoad` is now failing under Windows with Torch/GPU. I'm not sure what the intended behaviour is, but I would guess it's the test that is wrong, because the loaded tensor will be on some 'default' device, not current the `combo`.

PR #174, opened 2020-09-09
Documentation 1

gbaydin:
PR to track some documentation-related changes, expect more PRs to follow.


dsyme:
Looking good

dsyme:
@gbaydin I will work on docs in Tensor.fs if you like

Also would you like a Tensor.fsi to sepearate the API/docs from the implementation?  I can add that if you like.  It can make things simpler (and helps hide things) though causes some duplication

src/DiffSharp.Core/Extensions.fs
```diff
@@ -43,53 +43,53 @@ module Array =
 /// Contains extensions to the F# Seq module. 
 module Seq =
 
-    /// Get the index of the maximum element of the sequence
+    /// Gets the index of the maximum element of the sequence.
     let maxIndex seq =  seq |> Seq.mapi (fun i x -> i, x) |> Seq.maxBy snd |> fst
 
-    /// Get the index of the minimum element of the sequence
+    /// Gets the index of the minimum element of the sequence.
```
dsyme, line 49: .NET docs are usually written "Get", "Return"  etc. but I'm fine with this form.  You may have to keep an eye on me as it's a matter of habit to do the former :-)

dsyme:
> @gbaydin I will work on docs in Tensor.fs if you like


Actually I'm doing RawTensor first, just working my way through end to end

gbaydin:
> @gbaydin I will work on docs in Tensor.fs if you like
> 
> Also would you like a Tensor.fsi to sepearate the API/docs from the implementation? I can add that if you like. It can make things simpler (and helps hide things) though causes some duplication

Interesting option to do the documentation using .fsi files. I don't have a strong preference, I would like to leave this to your judgment. :)

gbaydin:
Merging this to make the new web site style available.

PR #176, opened 2020-09-09
make tensor type representation internal

dsyme:
The cases in the discriminated union should be hidden from the rest of the world so we can change it in the future, and control how the reverse/forward cases are generated

opUnary and opBinary need to be hidden because they are inlined.  This relates to #89 where we can provide public entry-points for user-defined extensions


dsyme:
@gbaydin I think we can just pull this - these union cases should be hidden going forward, since the API is used to access/create them

gbaydin:
@dsyme ok let's do it. 

Note: I think that the Tensor, TensorF, TensorR cases should be visible to the user and seeing the discriminated union cases in tensor's string representation was taking care of this in a simple way. This is so that they are aware of what their code is doing in debugging and prototyping. 

I don't know if making this internal hides this from the user. I think we can still give this information to the user by simply adding "F" and "R" suffixes somewhere in the Tensor's string representation (which I hope will be shown by default by fsi and dotnet interactive).

PR #177, opened 2020-09-09
Gpu test

VisualMelon:
Adds support for an environment variable `DIFFSHARP_TESTGPU` to control whether the test suite should use a cuda version of libtorch.

The idea is to make it easier to run the GPU tests locally (it's a one-time setup), and that this can be used by a future GPU supporting CI.


dsyme (approved):

dsyme:
@VisualMelon Thank you so much for this, it makes the GPU testing really easy!

PR #179, opened 2020-09-09
doc tooling update

dsyme:
@gbaydin  I started working on basic API docs for `Tensor`.  Could you check?  I think we will eventually move to needing a signature file.

I realised we needed a doc tooling update to handle the namespace docs properly

I'll work on this more later


dsyme:
I'll merge this, we can review and iterate on the docs in-situ

PR #182, opened 2020-09-10
Support -1 for expand

kevmal:
Assuming the goal is to match torch... This adds support for using `-1` with `expand` to specify the dimension size should remain unchanged.


gbaydin:
Thank you @kevmal. I revised this slightly to fit the library design better, merged dev, and also added an `expandAs` operation. Merging now. :)

PR #184, opened 2020-09-14
Placeholders for documentation

dsyme:
@gbaydin This adds about 500 placeholders for documentation. Not pretty but necesssary.  Seems best to add them first and then fill things in as it makes the codebase still look consistent 

I add `<summary>...</summary>` tags everywhere.  This is because I'd eventually expect most docs in this repo to use math `\(...\)` and `\[...\]` 

Let me know what you think

BTW 

* I worked on tooling to make F# IDE tools check docs as you type - it makes it so much easier to reliably write the docs https://github.com/dotnet/fsharp/pull/10118.  That will appear in upcoming editions of Visual Studio and F# support in VS Code

* I worked on making sure we can add examples too e.g. https://github.com/DiffSharp/DiffSharp/pull/184/files#diff-4db1087bc965bcae8b149b068e5b186cR14

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -4,6 +4,7 @@
 
      <LibTorchNugetVersion>1.5.6</LibTorchNugetVersion>
      <TorchSharpVersion>0.3.52267</TorchSharpVersion>
+	 <FSharpCoreVersion>4.7.2</FSharpCoreVersion>
       <!-- Standard nuget.org location -->
     <RestoreSources>https://api.nuget.org/v3/index.json</RestoreSources>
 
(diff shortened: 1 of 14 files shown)


dsyme:
Note: failure is just codecov/patch, we can ignore that

@gbaydin Are you ok with us merging this?  Best to get it in to avoid conflicts

dsyme:
@gbaydin I'll pull this now

gbaydin:
@dsyme thanks, I think it's OK and it was a good time to merge it. 

> I add <summary>...</summary> tags everywhere. This is because I'd eventually expect most docs in this repo to use math \(...\) and \[...\]

This makes sense.

> I worked on tooling to make F# IDE tools check docs as you type - it makes it so much easier to reliably write the docs dotnet/fsharp#10118. That will appear in upcoming editions of Visual Studio and F# support in VS Code

Looking forward to using it. Exciting improvements with documentation tools.

> I worked on making sure we can add examples too e.g. https://github.com/DiffSharp/DiffSharp/pull/184/files#diff-4db1087bc965bcae8b149b068e5b186cR14

Minor: I couldn't see the example there. Am I missing something?

PR #186, opened 2020-09-15
Fix save/load so it loads into a configuration, and doesn't store the device

### LINKED ISSUE #123: Tensor save/load issues
The testing work here https://github.com/xamarin/TorchSharp/issues/143 has raised some issues. For each we should compare to PyTorch

1. Should save/load on a GPU tensor bring it back as a GPU tensor?

2. Should the serialization format record the device, dtype and backend?  It currently does

3. Should the serialization format for TorchRawTensor be precisely the format used by Torch?  Currently DiffSharp uses .NET binary serialization.  We could use the save/load routines from Torch for tensors (see https://github.com/xamarin/TorchSharp/pull/148)

dsyme:
Fixes #123 

`dsharp.load` now takes parameters for dtype, device, backend etc.

The default dtype is the one in the tensor file

The default backend and device are the current ones

### DIFF
--- a/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
+++ b/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
@@ -58,7 +58,7 @@ type TorchRawTensor(tt: TorchTensor, shape: int[], dtype: Dtype, device: Device)
 
     inherit RawTensor()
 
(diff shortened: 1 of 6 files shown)


gbaydin:
@dsyme everything looks good, especially the tests. Also commented on #123

PR #187, opened 2020-09-15
update torchsharp version

dsyme:
This version of TorchSharp has getter/setters up to dimension 6, plus `t.RegisterAsMemoryPressure()` which we can use to help make memory management better when creating large stable tensors


gbaydin:
Looks good. :)

PR #188, opened 2020-09-16
Make underlying tensor handles available where necesssary

dsyme:
This makes available a handle to the underlying tensor objects where necessary, accessed via

        let t1 = dsharp.tensor([1.0f ; 1.0f ])
        t1.primalRaw.Handle 

etc.  For Torch backend the object is TorchTensor, for Reference backend it is the underlying .NET array

There are lots of reasons we may need this in practical debugging, performance investigations etc.


gbaydin:
@dsyme this looks good. 

I think the test for libtorch tensor handles is missing, but it's should be fine.

Issue #155, opened 2020-08-06
Reconsider both curried and tupled overloads on dsharp.add etc.


dsyme:
This pattern of having both curried and tupled methods is not recommended F# design and will likely cause problems (I noticed it in strange doc generation).  I think we will probably have to just choose one or another

```
    static member add(a:Tensor, b:Tensor) = a.add(b)
    static member add(b:Tensor) = fun (a:Tensor) -> a.add(b)
    static member sub(a:Tensor, b:Tensor) = a.sub(b)
    static member sub(b:Tensor) = fun (a:Tensor) -> a.sub(b)
    static member mul(a:Tensor, b:Tensor) = a.mul(b)
    static member mul(b:Tensor) = fun (a:Tensor) -> a.mul(b)
    static member div(a:Tensor, b:Tensor) = a.div(b)
    static member div(b:Tensor) = fun (a:Tensor) -> a.div(b)
```

gbaydin:
I would really like to keep this if possible.

As I'm writing DiffSharp code, supporting both styles gave me great flexibility. The curried ones work very well with functional F# style. For example here:

https://github.com/DiffSharp/DiffSharp/blob/33dd61eff9eb33460158c4eee98f744390402630/src/DiffSharp.Core/Distributions.fs#L202

But at the same time I am also able to use curried style for more PyTorch-like code. Is the problem only about the documentation?

dsyme:
> Is the problem only about the documentation?

There are multiple potential problems:

1. Order of arguments is confusing, e.g. `dsharp.sub x y` is `y - x` not `x - y`. That's really very confusing to people, and there's nothing to force currying to be used.

2. F# type inference and tooling is not really designed for APIs like this.  I just don't know if it will be that great. Certainly F# type inference just works much better if there are fewer overloads.  Also will parameter help work? I don't know.

3. 2x the documentation

4. It might be fine for div/add/mul/... etc. with two arguments but get whacky with other things that have more arguments. What's the general rule?

5. Intellisense will be more confusing to people - 2x the entries and they won't know which to choose

6. If we add shape inference, then we have 2x the API surface area

7. Too many ways to do the same thing

8. Are user-defined binary operators also meant to follow this pattern?

Of these (1) is to me too severe to embrace this on anything non-commutative.

There's commonly also a problem where things like this combine: a 2x multiplication of your API surface area becomes 4x then 8x.... when combined with something else (like overloading on scalars).

[...]

PR #191, opened 2020-09-21
Pipelined api

gbaydin:
This PR is to keep track of the curried/tupled API restructuring discussed in #155


src/DiffSharp.Core/DiffSharp.Compose.fs
```diff
@@ -0,0 +1,283 @@
+// Pipelined operations
+module DiffSharp.Compose
```
dsyme, line 1: Nice name

src/DiffSharp.Core/DiffSharp.Shorten.fs
```diff
@@ -0,0 +1,60 @@
+// Pipelined operations
+module DiffSharp.Shorten
```
dsyme, line 1: Maybe "DiffSharp.Abbreviations" though either is ok

dsyme (approved):

src/DiffSharp.Core/DiffSharp.Shorten.fs
gbaydin: I've been thinking about many names. At some point it was "Shorthand". Then I shortened the shorthand to this.

src/DiffSharp.Core/DiffSharp.Compose.fs
gbaydin: I like it too.

src/DiffSharp.Core/DiffSharp.Shorten.fs
gbaydin: I thought about "Abbreviations" too, but it felt too long. "Abbrev" would not be great either.

gbaydin:
@dsyme I'm going ahead and merging this for 1.0. We can revisit some choices post-1.0, but I'm quite happy with this design and compromise. Let's get this in and I will focus on the other things.

PR #192, opened 2020-09-23
Feature/nn2

gbaydin:
This is for keeping track of more NN work and examples


src/DiffSharp.Core/DiffSharp.Core.fsproj
```diff
@@ -19,5 +19,6 @@
     <Compile Include="Model.fs" />
     <Compile Include="Optim.fs" />
 	<PackageReference Update="FSharp.Core" Version="$(FSharpCoreVersion)" />
+  <PackageReference Include="SixLabors.ImageSharp" Version="1.0.1" />
```
VisualMelon, line 22: It may be preferable to put anything depending on a such an external dependency (i.e. substantial or commercial) in a separate assembly (e.g. `DiffSharp.Imaging` or something). It makes everything lighter and simplifies licensing for the consumer.

PR #195, opened 2020-09-28
Feature/nn3

gbaydin:
This is to keep track of more NN-related implementations.


dsyme:
@gbaydin I have resolved the small merge conflict in this work.  It was only a small conflict and you should pull it down and resolve against any other changes you have (or re-resolve the change yourself)

gbaydin:
This has been open for a while and has significant parts that are good to go. Merging now to continue new NN bits in `feature/nn4`

Issue #150, opened 2020-06-29
Zero size tensors


dsyme:
Torch supports tensors with size zero in one or more dimensions:

```
import torch
torch.Tensor().shape
torch.tensor(0).shape
torch.tensor([]).shape
torch.tensor([[]]).shape
torch.tensor([[]]).reshape(1,0).shape

```

gbaydin:
Note `tensor(0).shape` above is different from the others as it's not a "zero size tensor", it's just a scalar tensor (shape `[]`) with value 0 and it wouldn't be different from `tensor(0.2)` for example. I think it doesn't belong in this list.

The others in the list would have shape `[0]` representing an object of type tensor without any data or actual shape.

gbaydin:
The essence of this was implemented in #196, supporting the `tensor([])` case. I think this is fine for the time being as `tensor([[]])` variants aren't often seen in practice.

PR #196, opened 2020-09-29
Zero size tensors

gbaydin:
This addresses #150


dsyme (commented):
We need more tests for operations on zero sized tensors eg addition, sum etc.

gbaydin:
> We need more tests for operations on zero sized tensors eg addition, sum etc.

@dsyme Good point. I will look at it.

PR #198, opened 2020-09-29
Zero size tensors (2)

gbaydin:
This is following from #196 and it adds more tests to ensure behavior of zero-size tensors is correct in a range of operations.


gbaydin:
@dsyme I added more tests here which uncovered a couple of edge cases in the reference backend! I think this is good enough. Let me know what you think.

dsyme (approved):
looks good!

gbaydin:
Great, let's merge it!

PR #202, opened 2020-09-30
Add some docs

dsyme:
Continue to work through the docs in Tensor.fs

This also turns on "strict" doc checking where `fsdocs` will fail if the XML is not well formed


dsyme:
@gbaydin post-hoc review requested

PR #205, opened 2020-10-02
internalise backend dispatch on dtype

dsyme:
I'm doing some work on symbol execution for shapes, devices and dtypes.

This is a related structural simplification that removes the dispatch on dtype when finding the backend.  Instead the backend itself switches on dtype if it wishes too (it doesn't have to, if the underlying tensor implementation can just accept the dtype as a parameter).

This makes the API to the backends a little simpler - only one specific type needs to be implemented to constitute a backend.  

There's no other change in this PR


dsyme:
@gbaydin I'll merge this. Let me know if you need to discuss

gbaydin:
Thanks, I will check.

PR #208, opened 2020-10-06
switch to CheckEqual in testing

dsyme:
@gbaydin This changes the testing to use a `CheckEqual(T, T)` which requires that the source and target have the same type.  The actual equality still uses AreEqual 

The fact we were using an `AreEqual(obj,obj)` overload has been completely biting me when I'm working with shapes, as it's really easy to get the types wrong on one of the arguments.  This new approach is much more robust under change


dsyme:
@gbaydin I'll just merge this when it's green, it should be easy to integrate into any work you have outstanding I think

PR #215, opened 2020-10-10
split out Tensor.Slicing.fs

dsyme:
Tensor.fs was getting a bit large - this splits out the generated slicing code


dsyme:
@gbaydin Post-hoc review requested, I think this is ok.  The one downside is that the specific slicing overloads can't be used in Tensor.fs but we're not using them currently, so I'm guessing you're ok with this.

Among other things this makes editing inside Tensor.fs much more pleasant since the edit-error-refresh cycle is much snappier, as is autocomplete etc.

gbaydin:
Sounds good especially if it helps with improving the development cycle.

PR #219, opened 2020-10-15
Document one API to try things out

migueldeicaza:


dsyme:
Cool

Just to say the preview versions of the latest upcoming release of Visual F# Tools in Visual Studio for WIndows has XML doc checking built in

Not yet available in VS for Mac - the F# tooling there will need to be updated to use the latest F# Compiler Service version to pick that up.

dsyme:
> Just to say the preview versions of the latest upcoming release of Visual F# Tools in Visual Studio for WIndows has XML doc checking built in

Because of this productivity can be a bit trashed when not using this tooling.  

I'm working with the community to get the tooling updated in Ionide in VSCode at least, which would give a viable contribution path for the docs on Mac.

thanks

src/DiffSharp.Core/DiffSharp.Compose.fs
```diff
@@ -8,7 +8,17 @@ module DiffSharp.Compose
 // static member sub(b:Tensor) = fun (a:Tensor) -> a.sub(b)
 
 type dsharp with
-    /// <summary>TBD</summary>
+    /// <summary>Returns a tensor where each row contains <paramref name="numSamples"/> indices sampled from the multinomial probability distribution located in the corresponding row of tensor input.</summary>
```
dsyme, line 11: In general I haven't been using `paramref` , partly because we don't generate any links for this.  There is checking for this in the latest F# tools to double check the reference at least, but tbh it's so clumsy to type it in I think we can use just `<c>... </c>` or plan text for now and take a separate pass to add paramref later

dsyme: Note 

* F# tools don't yet have documentation-aware rename-refactor for paramred and other references.  This makes cross-references of any kind a bit hard.  

* Only paramref are checked, other cross-references are not checked.

dsyme:
@migueldeicaza A few more things

* There are multiple entry points to the same functionality, as in pytorch, e.g. `a.conv1d(b)` and `dsharp.conv1d(a,b)` as well as the Compose and other entry points - this is intended (@gbaydin's call, it makes sense for pytorch ergonomics with an F# angle too).

* Most functions are documented in `DiffSharp.fs` .  You can just copy across.

PR #222, opened 2020-10-15
docs for models and namespaces

dsyme:
Adds docs for the remaining namespaces and some docs on the Model types

@gbaydin please review - the docs are taken from PyTorch as before.  We could also jsut use links across to the PyTorch docs instead of have the repeated `<remarks>` sections, let me know what you'd like.

### DIFF
--- a/src/DiffSharp.Core/Data.fs
+++ b/src/DiffSharp.Core/Data.fs
@@ -8,6 +8,11 @@ open System.Net
 open System.IO
 open System.IO.Compression
 
+/// <namespacedoc>
+///   <summary>Contains data sets and components related to data loading.</summary>
+/// </namespacedoc>
+///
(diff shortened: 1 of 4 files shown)


gbaydin:
Hi @dsyme I think we shouldn't take documentation from PyTorch.

- We're not wrapping PyTorch models or porting torch.nn module code in DiffSharp. Copying documentation can give this wrong impression to the users.
- PyTorch is used as a very good and stable guidance for the API people are used to, but this is where the correspondence ends. We build model infrastructure on Tensors completely independently.
- Here's a little example of how nn implementations can differ in practice: https://github.com/DiffSharp/DiffSharp/blob/830c90961e575c23c105d505988df5460ffdd84e/src/DiffSharp.Core/Model.fs#L515 

So please let's minimize this for future document writing.

PR #230, opened 2020-11-03
Redo #88 batch matmul

dsyme:
This is a redo of https://github.com/DiffSharp/DiffSharp/pull/88

A bad commit crept into the final merge  of #88, see https://github.com/DiffSharp/DiffSharp/pull/88/commits/3c94672c1abd19229385738458986864a48ee614, which completely rewrote Tensor.fs (encoding, or line endings or something).  I've checked right back in history and whatever previous formats we were using for Tensor.fs do not match this.


gbaydin:
@dsyme sorry about that. Was this my mistake somehow?

dsyme:
> @dsyme sorry about that. Was this my mistake somehow?

Not sure, not a problem either way

PR #231, opened 2020-11-03
Mutable raw tensors for internal use

dsyme:
This puts in place the backend foundations for using in-place mutation on a separate type of MutableRawTensor, primarily for internal use inside DiffSharp.  This will certainly be needed in the near future for obvious performance improvements, which is why I want to get this in place now.

A separate type is used to allow us to carefully isolate out and reason about any code that uses mutation

TorchSharp https://github.com/xamarin/TorchSharp/pull/171, https://github.com/xamarin/TorchSharp/pull/172 and https://github.com/xamarin/TorchSharp/pull/173 contain various additions to help complete the set of inplace operations available.

The changes are pretty clear, I think. Naming on MutableRawTensor could be improved once we start writing code to utilise this


dsyme:
Discussed with @gbaydin, merging as this is foundational for future work

PR #235, opened 2020-11-06
cleanup and assert shapes in reverse mode

dsyme:
* `abstract member` --> `abstract`

* Add an assertion to check shapes of adjoints in reverse mode

* Fix a mistake in the slicing shape computation where the `1` flag was being set for slices like `t.[3..]` resulting in incorrect result shapes when using that notaion to slice the first or last single element.  

* Fix a mistake in boundsToShape, which is used in two slightly different ways - one to get the result shape of slicing, the other to push adjoints in the SliceTT reverse mode op back into the adjoint for the originating tensor.

* Rename the union case `Tensor` to `Tensor0`. This is done because I'm fairly certain that some future changes (either performance or opt-in mutability in tensors) will require us to adjust the `Tensor` type representation details, e.g. to be a class, or record, or similar. To do this accurately with fewer code changes we need this union case name to be something besides the same name as the enclosing type.  One example of this is in [this branch](https://github.com/DiffSharp/DiffSharp/compare/feature/mut-backprop), but it's also a general issue about making it easier to make future possible changes to the internal representation of this type.  

  So I don't mind what the name of the union case is as long as it's not `Tensor`.... :-) @gbaydin is this ok?


dsyme:
@gbaydin I will merge this, see rationale above for the reasons for changing the name of the union case.

If you'd like it to be Tensor1 or TensorC (for constant) that's cool, just not Tensor, I can explain in more detail why if you like

src/DiffSharp.Core/Tensor.fs
```diff
@@ -2796,6 +2796,10 @@ type Tensor =
     /// <summary>Push the given value as part of the reverse-mode computation at the given output tensor.</summary>
     /// <param name="value">The value to apply.</param>
     member t.reversePush(value:Tensor) =
+        let check (v:Tensor,t:Tensor) = 
+            // check the shapes of the adjoints match the nodes to which they are being propagated
+            assert (Shape.canExpand v.shape t.derivative.shape || Shape.canExpand t.derivative.shape v.shape)
```
gbaydin, line 2801: This is making me extremely worried. Let's chat about it when we meet.

dsyme: Cool thanks for checking post hoc

The assert isn't failing for any tests we have, do you think it's too tight or too loose?

Yes let's discuss when we meet

PR #236, opened 2020-11-09
Adding benchmarking for micro operations

dsyme:
This starts to add systematic benchmarking of micro operations at the different layers of tensors to DiffSharp.Benchmarks

It also fixes tensor creation direct from .NET array data to be much faster


dsyme:
Also bumps to TorchSharp v0.3.52293

dsyme:
I would like to augment this with equivalent python too.  I'll likely work on that tomorrow.

PR #243, opened 2020-11-13
Scalar tensors as IConvertible

dsyme:
This was removed in #242 but I'm adding it back in here, it's ok to have this.  Iv'e added the extra testing and GetTypeCode implementation mentioned there.


gbaydin:
I'm glad IConvertible is back.

PR #248, opened 2020-11-19
Feature: Allow generic In/Out on models

dsyme:
Generalises models to allow a generic In/Out, as used in transformer architectures

    type Model<'In, 'Out>
    type Model = Model<Tensor, Tensor>

There are other ways to do this but this seems as good as any.

### DIFF
--- a/src/DiffSharp.Core/Model.fs
+++ b/src/DiffSharp.Core/Model.fs
@@ -145,7 +145,7 @@ type Mode =
 
 /// <summary>Represents a model, primarily a collection of named parameters and sub-models and a function governed by them.</summary>
 [<AbstractClass>]
-type Model() =
+type BaseModel() =
     [<DefaultValue>]
     val mutable mode: Mode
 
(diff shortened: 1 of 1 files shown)


dsyme:
@gbaydin I think this can be merged, or you can take the change and apply it into your own work on transformers?

It raises separate questions of generalizing differentiable objects to beyond tensors but I think we can deal with those later, this seems like a reasonable minimal start?

gbaydin:
Actually this is indeed very relevant for the type of things I'm working on. For example in recurrent models I had to think about a design to pass and return hidden state tensors in addition to the input and return values of `.forward`. Thanks for pinging, I'm going to have a close look and merge.

gbaydin:
@dsyme I added a couple of fixes. 

I think the most important one was that `Model<'In, 'Out>.clone` was returning `Model<Tensor, Tensor>` which I fixed to return `Model<'In, 'Out>`.

I also moved a few things that were not involving `<'In, 'Out>` to `BaseModel`.

I'm really happy about this PR in general. I think this is quite nice and powerful! For example, we can do things like:

* `Model<string, string>` for machine translation between languages
*  `Model<unit,Tensor>` for some generative models etc.

Also after reading through all `<'In, 'Out>`s this reminds me of https://www.in-n-out.com/ :)

dsyme:
This is great, thanks for refining it!

PR #250, opened 2020-11-20
bump .NET SDK version

dsyme:
Bump so we can use F# 5.0 features (.NET 5.0 SDK is now requried)

Looks like we also requires a .NET 3.1 SDK to actually run tests

### DIFF
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,6 +16,10 @@ jobs:
     - name: Setup dotnet
       uses: actions/setup-dotnet@v1
       with:
+        dotnet-version: 3.1.301
+    - name: Setup dotnet
+      uses: actions/setup-dotnet@v1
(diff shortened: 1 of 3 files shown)


gbaydin:
Hi @dsyme do you know why we need dotnet 3.1 for the tests?

dsyme:
> Hi @dsyme do you know why we need dotnet 3.1 for the tests?

The tests run as netcoreapp3.1, I think the CI installation technique doesn't automatically install the runtime for netcoreapp3.1.  

I tried net5 when I did this but hit some issue - it might have BenchmarkDotnet, I'm not sure.

We should be able to move to purely net5 soon I think.  Likely best to wait a month or two while other tools adjust

gbaydin:
Ok, I could also test this within a Docker environment to inspect what's going on.

PR #251, opened 2020-11-25
Factor out some common code

dsyme:
Share the repeated code fragments for combining stride/strides and other argument pairs


gbaydin:
I'm really glad we have all the tests in place that makes me feel easy about this change. :)

PR #256, opened 2020-11-27
allow empty parameter dictionary

dsyme:
Currently `.parameters` on a model fails if there are no parameters.  This returns an empty tensor for this case


dsyme:
@gbaydin this is a simple fix, I presume you'll find it ok, please posthoc review

gbaydin:
This looks fine.

PR #252, opened 2020-11-25
feature/avgpool - avgpool1d, avgpool2d, avgpool3d

dsyme:
Adds avgpool support

@gbaydin this uses the updated #89 to stop polluting Tensor,fs, and to have `DiffSharp.AvgPool.fs` extend all the necessary API surfacings all at once

PR #260, opened 2020-11-28
[WIP] feature/activations

dsyme:
This puts in place the outline of the activation functions necessary for the range of models I've been looking at porting to DiffSharp.

- Like #252 this uses (internall) the extension API of #89 so we don't have to shove everything into Tensor.fs.  It can be done without but then sooner or later we have to refactor.

- The implementations in Reference.RawTensor.fs and derivatives DiffSharp.Activations.fs in still need to be added

- Testing is needed

LibTorch provides primitives for the reverse mode operations though these are not yet available, and we would also need to the 2nd derivatives if we utilise those.

Cheers
don

### DIFF
--- a/src/DiffSharp.Backends.Reference/Reference.RawTensor.fs
+++ b/src/DiffSharp.Backends.Reference/Reference.RawTensor.fs
@@ -349,6 +349,13 @@ type RawTensorCPU<'T when 'T : equality and 'T :> scalar>(values: 'T[], shape: S
     override t.RoundInPlace() = t.SetValues <| t.RoundT()
     override t.AbsInPlace() = t.SetValues <| t.AbsT()
     override t.ReluInPlace() = t.SetValues <| t.ReluT()
+    override t.LeakyReluInPlace(negativeSlope) = t.SetValues <| t.LeakyReluT(negativeSlope)
+    override t.EluInPlace(alpha, scale, inputScale) = t.SetValues <| t.EluT(alpha, scale, inputScale)
+    override t.GeluInPlace() = t.SetValues <| t.GeluT()
+    override t.HardsigmoidInPlace() = t.SetValues <| t.HardsigmoidT()
+    override t.HardswishInPlace() = t.SetValues <| t.HardswishT()
+    override t.Relu6InPlace() = t.SetValues <| t.Relu6T()
+    override t.SiluInPlace() = t.SetValues <| t.SiluT()
     override t.SoftplusInPlace() = t.SetValues <| t.SoftplusT()
     override t.SigmoidInPlace() = t.SetValues <| t.SigmoidT()
     override t.ExpInPlace() = t.SetValues <| t.ExpT()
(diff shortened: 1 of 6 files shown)


src/DiffSharp.Core/DiffSharp.Activations.fs
```diff
@@ -0,0 +1,108 @@
...
+
+            /// <summary>Applies the leaky rectified linear unit function element-wise</summary>
+            /// <remarks>\[\text{LeakyReLU}(x) = \max(0, x) + \text{negative\_slope} * \min(0, x)\]</remarks>
+            /// <param name="negativeSlope">Controls the angle of the negative slope. Default: 0.01.</param>
+            member a.leakyRelu(?negativeSlope:float) =
+                let negativeSlope = defaultArg negativeSlope 0.01
+                let zeros = a.zerosLike() in zeros.max(a) + negativeSlope * zeros.min(a)
+
+            /// <summary>Applies the rectified linear unit function element-wise.</summary>
+            member a.elu() = elementwiseOp (fun a -> a.EluT()) (fun t a td -> failwith "deriv of elu NYI") a
```
dsyme, line 27: derivatives needed here

src/DiffSharp.Backends.Reference/Reference.RawTensor.fs
```diff
@@ -825,6 +832,41 @@ module internal RawTensorCPU =
         let result = t.Values |> Array.map (max zero< ^T >) 
         (result, t.Shape)
 
+    let inline LeakyReluT(t: RawTensorCPU< ^T >, negativeSlope: double) : (^T[] * Shape) =
+        failwith "tbd"
+        //let result = t.Values |> Array.map (max zero< ^T >) 
+        //(result, t.Shape)
+
+    let inline EluT(t: RawTensorCPU< ^T >) : (^T[] * Shape) =
+        failwith "tbd"
```
dsyme, line 841: reference function implementations needed here

dsyme:
Having coded up the outline of this, I feel we should change the backend API to make this need less plumbing.  

For example for elementwise-functions we should just have 

```fsharp
type RawTensor() =
    abstract Apply: op: ElementwiseOperation -> RawTensor
    abstract ApplyInPlace: op: ElementwiseOperation -> unit

type ElementwiseOp = 
    | Relu 
    | LeakyRelu of negativeSlope: double
    | Gelu 
    ...
```

and so on.  Then everything becomes a bit simpler.  It also feels like backend should optionally implement these, rather than requiring all backends to always implement them, e.g. 

```fsharp
type BackendTensorStatics() = 
    ...
    /// Checks if a backend supports an optimized entry point for an operation
    abstract SupportsOperation: ElementwiseOperation -> bool
```

In the cases where an optimized entry point is not available then differentiable Tensor code would be used instead.  Then the reference backend would simply not need to implement these at all, and we would just dynamically test if the backend supports the operation.

gbaydin:
In principle I'm not a fan of treating "activations" as special entities in DiffSharp. They are just some differentiable functions and there is nothing different between `tensor.relu` or `tensor.exp` (exp actually gets used as activation sometimes). In practice anything can be an "activation". So I think there can't be a clear and logical rule for deciding which functions should go into the proposed `DiffSharp.Activations.fs` file.

> LibTorch provides primitives for the reverse mode operations though these are not yet available, and we would also need to the 2nd derivatives if we utilise those.
I don't understand the part about the "2nd derivatives". This wouldn't apply to DiffSharp because 2nd or any other derivatives would be just spontaneously available when we implement the base function. Are you referring to adding "reverse activations" as primal ops in DiffSharp?

dsyme:
> Please do not merge until I check everything starting with #89

Yes I'm not going to merge any `feature/*` work until you've reviewed thoroughly and approved.

> In principle I'm not a fan of treating "activations" as special entities in DiffSharp. 

Yeah, I understand, both terminology-wise and in terms of how to implement these functions

On terminology - I agree about "activations" - I didn't really mean to introduce the name into DiffSharp.  At the moment the thing is all AutoOpen (only `open DiffSharp` is needed) so we can use whatever name for the module we like - the word "activation" isn't needed in the DiffSharp user model.

On implementation - yes the only valid reason not to just write out the function as DiffSharp Tensor code is perf.   These are important for performance in multiple models.  And it's clear to me that DiffSharp code that copies tensors 5 or 6 times to compute, say, `a.silu()` - doing nothing in-place - is going to be many, many times slower than the underlying LibTorch primitive.  We can get exact figures but I don't expect it will look good - I'd expect 10x slower, maybe much more. To be honest this is a major concern for me in the DiffSharp architecture v. PyTorch  - we will repeatedly have functions that are very fast in PyTorch and their equivelents in DiffSharp will be slow.  

> I don't understand the part about the "2nd derivatives". This wouldn't apply to DiffSharp because 2nd or any other derivatives would be just spontaneously available when we implement the base function. Are you referring to adding "reverse activations" as primal ops in DiffSharp?

Yes, adding reverse activations to TorchSharp and RawTensor gives us reverse ops (effectively the first derivatives), but those ops in turn need their own reverse ops (based on the 2nd derivatives), which do not have corresponding LibTorch  functions.  At some point this needs to be closed and this can only be done with a DiffSharp Tensor expression.  We do that already, e.g. at the moment we have

```fsharp
    member a.relu() = 
      elementwiseOp 
          (fun a -> a.ReluT()) 
          (fun _t a td -> let sap = a.sign() in td * sap.abs() * (sap + 1.) / 2.) a
```
(that derivative code comes from Tensor.fs)

Now if ReluReverse is in the RawTensor ops then we can add the reverse op:

```fsharp
    member a.relu() = 
      elementwiseOp 
          (fun a -> a.ReluT()) 
          (fun _t a td -> a.reluReverse()) 
          a

    member internal a.reluReverse() = 
      elementwiseOp 
          (fun a -> a.ReluReverseT()) 
          (fun _t a td -> (* ***what goes here?** *)) 
          a
```

Howeve the expression that goes in the "what goes here" spot has to be the reverse-of-the-reverse (which happens to be `a.zerosLike()` in the case of Relu but you get what I mean).

Again, the only reason we would go down this path is to pick up the performance gains of ReluT and ReluReverseT

gbaydin:
Thanks for the clarification. I was guessing this is due to performance considerations and I understand and support this. Let's make sure we get the design right and we should of course rely on the backends as much as we can for commonly used functions such as these "activations".

Actually this was also my way of thinking when implementing things like relu, leakyrelu, etc. I was seeing them as placeholders until we do this type of PR. My plan was to:
1. Implement these functions as compositions of other DiffSharp ops to incorporate the necessary operations into the code base, implement the tests, etc. to have a baseline implementation.
2. Reimplement some parts of the ops to rely on backend.

> (fun _t a td -> (* ***what goes here?** *)) 

We should be able to derive the necessary things analytically for such cases.

dsyme:
OK cool. Tricky to find the right balance of methodology, with about 1000 optimised routines in libtorch.

Another methodology is this

1. Keep pushing things into RawTensor, perhaps unused (like the mutability support) or unimplemented for Reference

2. Make educated guesses about which ops are performance critical and give these first class treatment

> We should be able to derive the necessary things analytically for such cases.

> `(fun _t a td -> (* *what goes here? *))`

I was wondering if it's legitimate to write something like this, for `let gelu x = some-expr`

    member a.gelu() = 
      elementwiseOp 
          (fun a -> a.GeluT()) 
          (fun _t a td -> a.geluReverse()) 
          a

    member internal a.geluReverse() = 
      elementwiseOp 
          (fun a -> a.GeluReverseT()) 
          (fun _t a td -> dsharp.diff2 (fun x -> some-expr) a * td )

That is invoke `dsharp.diff2` (or some other differentiation routines, knowing this is an element-wise operation) on the expression for the original analytical function - in the middle of providing the 2nd reverse.  This would save us deriving the derivatives analytically?

Issue #258, opened 2020-11-28
Review of merged #235 and #262


gbaydin:
@dsyme this merge #235 is slightly worrying. Let's discuss when we meet next time.

About the `Tensor0` case: I understand that you have some very good reason for this based on how you feel the codebase will evolve. I don't like the change, but I have a feeling it needs to happen. `Tensor0` is not a great name. Let's think about a better one `TensorC` can be better. What about `Tensor_`?

Another thing is the "%A" printing which has been relying on the default DU printer. So that we got it look like `Tensor [1.; 2.; 3.]` I have a feeling this can be addressed through other means. (Later edit: I can see now that there is a new tensor printing code merged here #234, will check.)

**More importantly** I have significant worries about the following two changes. I have a feeling that the mistakes you identified were not mistakes but were some intentional design influenced by Python-like conventions. I need to remember and inspect the behavior and then try to understand your change better.

> Fix a mistake in the slicing shape computation where the 1 flag was being set for slices like t.[3..] resulting in incorrect result shapes when using that notaion to slice the first or last single element.

> Fix a mistake in boundsToShape, which is used in two slightly different ways - one to get the result shape of slicing, the other to push adjoints in the SliceTT reverse mode op back into the adjoint for the originating tensor.

Later edit: I think this is also connected with why you added the `check` here
```fsharp
member t.reversePush(value:Tensor) =
    let check (v:Tensor,t:Tensor) = 
        // check the shapes of the adjoints match the nodes to which they are being propagated
        assert (Shape.canExpand v.shape t.derivative.shape || Shape.canExpand t.derivative.shape v.shape)
```

I remember there were some very clear reasons and a careful design for the behavior you changed and it was also about the reverse mode shapes. I'm quite concerned about this merged code and I think the `expand` check is not a good idea. We don't want things in reverse mode to rely on broadcasting/expansions etc. We need to check if the reverse pushed value has exactly the same shape with the primal of the node to which we are propagating. Anything other than that can be a cause for silent erroneous behavior with derivatives in future code. I think it's very likely I would like to revert this merge.

dsyme:
@gbaydin We should talk about this.

> We don't want things in reverse mode to rely on broadcasting/expansions etc. 

Right, I 100% agree, but the code (before and after #235) does rely on broadcasting/expansions I'm afraid. In particular there are places it uses "dsharp.zero()" for the adjoint or derivatives.  

[...]

PR #262, opened 2020-11-28
Tighten assert in reverse

dsyme:
Address issues raised about #235 in review #258

* add test that shows slicing fix in #235 fixes a real problem (I've verified TestDerivativeSliceTFinalColumn failed before #235)

* add test that shows that `boundsToShape` change in #235 was not necessary (it was actually a no-op brought over mistakenly from #237), and revert that

* show we can tighten the assert added in #235 per discussion in #258 to prove there's no strange broadcasting going on.

### DIFF
--- a/tests/DiffSharp.Tests/TestDerivatives.fs
+++ b/tests/DiffSharp.Tests/TestDerivatives.fs
@@ -2410,6 +2410,21 @@ type TestDerivatives () =
             Assert.True(revyd.allclose(revydCorrect, 0.01))
 
     [<Test>]
+    member _.TestDerivativeSliceT_2D () =
+        for combo in Combos.AllDevicesAndBackendsFloat32 do
+
+            let revx = combo.tensor([[54.7919; 70.6440; 16.0868; 74.5486; 82.9318]; 
+                                     [54.7919; 70.6440; 16.0868; 74.5486; 82.9318]]).reverseDiff()
+            let revz = revx.[*,2..2]
+            let revzCorrect = combo.tensor([16.0868; 16.0868])
+            revz.reverse(combo.tensor([0.9360;0.9360]))
+            let revxd = revx.derivative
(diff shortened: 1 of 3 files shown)


tests/DiffSharp.Tests/TestDerivatives.fs
```diff
@@ -2354,6 +2354,20 @@ type TestDerivatives () =
             Assert.True(revz.allclose(revzCorrect, 0.01))
             Assert.True(revxd.allclose(revxdCorrect, 0.01))
 
+
+    [<Test>]
+    member _.TestDerivativeSliceTFinalColumn () =
+        for combo in Combos.AllDevicesAndBackendsFloat32 do
+            let revx = combo.tensor([54.7919; 70.6440; 16.0868]).reverseDiff()
```
gbaydin, line 2361: We need the forward derivative test

tests/DiffSharp.Tests/TestDerivatives.fs
```diff
@@ -2409,6 +2423,21 @@ type TestDerivatives () =
             Assert.True(revxd.allclose(revxdCorrect, 0.01))
             Assert.True(revyd.allclose(revydCorrect, 0.01))
 
+    [<Test>]
+    member _.TestDerivativeSliceT_2D () =
+        for combo in Combos.AllDevicesAndBackendsFloat32 do
```
gbaydin, line 2428: We need the forward derivative test

tests/DiffSharp.Tests/TestDerivatives.fs
gbaydin: We also need at least one test for >1 dimensional case

dsyme:
Cool I'll add systematic slicing testing tomorrow

gbaydin:
> Cool I'll add systematic slicing testing tomorrow

I mean I think just a 2d case would suffice to convince us that the implementation generalizes to things > 1d. I didn't want to propose spending too much time with this.

PR #266, opened 2020-11-30
Update testing dependencies, target net5.0 (remove netcoreapp3.1)

gbaydin:
@dsyme this is doing the following changes:

- Update `netstandard2.1` and `netcoreapp3.1` to `net5.0` as suggested [here](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#suggested-targets)
- Update the `NUnit3TestAdapter` and `Microsoft.NET.Test.Sdk` to their latest versions in dotnet 5 F# templates. I copied the versions given by a new test project one gets by `dotnet new nunit -lang "F#"`
- Updates the dockerfile to dotnet 5, which I used to test these would work in a system without dotnet core 3.1 installed.
- Updates the CI runners to remove dotnet 3.1 installation step

This works for me in a Docker container with only dotnet 5 sdk in it:
```
docker run -it --rm diffsharp
root@38f8de59415c:/code/DiffSharp# dotnet test
  Determining projects to restore...
  All projects are up-to-date for restore.
  DiffSharp.Core -> /code/DiffSharp/src/DiffSharp.Core/bin/Debug/net5.0/DiffSharp.Core.dll
  DiffSharp.Backends.TestDuplicate -> /code/DiffSharp/tests/DiffSharp.Backends.TestDuplicate/bin/Debug/net5.0/DiffSharp.Backends.TestDuplicate.dll
  DiffSharp.Backends.Torch -> /code/DiffSharp/src/DiffSharp.Backends.Torch/bin/Debug/net5.0/DiffSharp.Backends.Torch.dll
  DiffSharp.Backends.Reference -> /code/DiffSharp/src/DiffSharp.Backends.Reference/bin/Debug/net5.0/DiffSharp.Backends.Reference.dll
  DiffSharp.Tests -> /code/DiffSharp/tests/DiffSharp.Tests/bin/Debug/net5.0/DiffSharp.Tests.dll
Test run for /code/DiffSharp/tests/DiffSharp.Tests/bin/Debug/net5.0/DiffSharp.Tests.dll (.NETCoreApp,Version=v5.0)
Microsoft (R) Test Execution Command Line Tool Version 16.8.0
Copyright (c) Microsoft Corporation.  All rights reserved.

Starting test execution, please wait...
A total of 1 test files matched the specified pattern.

Passed!  - Failed:     0, Passed:   379, Skipped:     0, Total:   379, Duration: 30 s - /code/DiffSharp/tests/DiffSharp.Tests/bin/Debug/net5.0/DiffSharp.Tests.dll (net5.0)
root@38f8de59415c:/code/DiffSharp# exit
```

Could you please have a look and see if there are any problems?

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "fsharp.formatting.commandtool": {
-      "version": "7.2.9",
+      "version": "8.0.0",
       "commands": [
         "fsdocs"
       ]
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -16,10 +16,6 @@ jobs:
     - name: Setup dotnet
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.301
-    - name: Setup dotnet
-      uses: actions/setup-dotnet@v1
-      with:
         dotnet-version: 5.0.100
     - name: Install dependencies
       run: dotnet restore
--- a/.github/workflows/pull-request-debug.yml
+++ b/.github/workflows/pull-request-debug.yml
@@ -15,10 +15,6 @@ jobs:
     - name: Setup dotnet
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.301
-    - name: Setup dotnet
-      uses: actions/setup-dotnet@v1
-      with:
         dotnet-version: 5.0.100
     - name: Install dependencies
       run: dotnet restore
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -15,10 +15,6 @@ jobs:
     - name: Setup dotnet
       uses: actions/setup-dotnet@v1
       with:
-        dotnet-version: 3.1.301
-    - name: Setup dotnet
-      uses: actions/setup-dotnet@v1
-      with:
         dotnet-version: 5.0.100
     - name: Install dependencies
       run: dotnet restore
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,4 @@
-FROM mcr.microsoft.com/dotnet/core/sdk:3.1
+FROM mcr.microsoft.com/dotnet/sdk:5.0
 WORKDIR /code/DiffSharp
 COPY . /code/DiffSharp
 RUN dotnet build
--- a/bundles/DiffSharp-cpu/DiffSharp-cpu.fsproj
+++ b/bundles/DiffSharp-cpu/DiffSharp-cpu.fsproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
   
   <ItemGroup>
--- a/bundles/DiffSharp-cuda-linux/DiffSharp-cuda-linux.fsproj
+++ b/bundles/DiffSharp-cuda-linux/DiffSharp-cuda-linux.fsproj
@@ -1,7 +1,7 @@
 ﻿<Project Sdk="Microsoft.NET.Sdk">
 
   <PropertyGroup>
-    <TargetFramework>netcoreapp3.1</TargetFramework>
+    <TargetFramework>net5.0</TargetFramework>
   </PropertyGroup>
   <ItemGroup>
     <Compile Include="Empty.fs" />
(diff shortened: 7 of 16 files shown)


gbaydin:
Ok everything works except fsdocs. It looks like fsdocs needs to be updated to work without requiring net core 3.1.

image

VisualMelon:
What is the rationale for removing netcore 3.1 and netstandard 2.1 support in the back-ends and core library? Both will be fully supported for a couple of years yet (which net5.0 will not); are there net5.0 features/libraries expected to become dependencies?

gbaydin:
One motivation is to not ask and users/developers to install two separate dotnet versions (3.1 and 5.0) just to run tests and recreate documentation as it is currently the case. Actually I'm confused about this and I don't know why having 5.0 sdk was not enough to run a `netcoreapp3.1` test application as we observed. The table [here](https://docs.microsoft.com/en-us/dotnet/standard/frameworks#net-5-os-specific-tfms) says net5.0 is compatible with `netcoreapp3.1`.

The other motivation is the information in many places ([one example](https://visualstudiomagazine.com/articles/2020/09/16/net-standard-future.aspx)) saying `net5.0` is now the suggested target for cross-platform libraries. I like the simplicity this brings and I'm fine with saying you need dotnet 5 to use DiffSharp.

If there is an easy way for supporting both (without requiring both), I'm fine with that too. I think it should be possible to target multiple frameworks like `<TargetFrameworks>netcoreapp3.1;net5.0</TargetFrameworks>`. I will leave the choice here to @dsyme.

cgravill:
The fsdocs failure might be down to roll forward policy: https://docs.microsoft.com/en-us/dotnet/core/versions/selection#framework-dependent-apps-roll-forward

VisualMelon:
Ease of dev is a reasonable reason. I would suggest dual-targeting the consumer-facing stuff, though, so long as there are no dependencies that make that a problem. (You should be able to multi-target netstandard2.1 and netcoreapp3.1 without having netcoreapp3.1 installed).

<hr />

Other notes:

The net 5.0 runtime can execute a netcore3.1 application, but only if the runtime config asks for it to do so. By default, the runtime config for an application compiled against netcore3.1 will request netcore3.1.0, and (as in cgravill's link) it will be unhappy if you only have net5.0 installed.

You don't need to target net 5.0 to use F#5 (just be compiling with the .net 5.0 sdk), but I don't know whether there are features that will require a net 5.0 target (my understanding is that it should be good for anything >= netstandard2.0, but I could be mistaken). That is, it should be possible to use F#5 in the library code-base without requiring the consumer to use net5.0 for their own work.

dsyme:
This is a long, long rabbit hole to go down.

1. We need to use F# 5, since it has features we need (e.g. `open type`, string interpolation, nullable interop).

2. Using F# 5 requries an install of .NET 5 SDK, which only installs .NET 5 Runtime by default.

3. The fsdocs tool requires a .NET Core 3.x runtime and it appears running fsdocs on .NET 5 Runtime is a total PITA.  I've recorded this issue here: https://github.com/fsprojects/FSharp.Formatting/issues/620. It seems so simple, but resolving this seems to require major work in fsdocs repository and also involves updating both FAKE and paket to be net5.0-aware versions.  I've tried doing that but there are a whole series of obscure errors along the way

I'm annoyed this is not simple.   For now I recommend we just stick to what we have until fsdocs is sorted out.

cgravill:
Is it possible to adjust the fsdocs call?

https://github.com/DiffSharp/DiffSharp/blob/12137670a9aa347a0ac3199b45526c4118a821a7/.github/workflows/pull-request.yml#L31-L32

If we were doing a `dotnet XYZ.dll` I understand it could be adjusted to `dotnet --roll-forward LatestMajor XYZ.dll` like this

with a runtime config:
https://docs.microsoft.com/en-us/dotnet/core/tools/dotnet#runtime-options

but that doesn't seem to be supported with .NET Tools call. Perhaps we could still use the `DOTNET_ROLL_FORWARD` environment variable? All this assumes that fsdocs would _just run_ on .NET 5.0

dsyme:
https://github.com/fsprojects/FSharp.Formatting/pull/621 should allow fsdocs to roll forward

dsyme:
It turns out running fsdocs (and any F# tool based on FSharp.Compiler.Service which processes scripts) on .NET Core 5.0 requrires [this core F# fix](https://github.com/dotnet/fsharp/pull/10575) and an update to FCS

So this will take a little while to come through the pipe

dsyme:
I updated to fsdocs 8.0.0 which will hopefully allow us to use .NET 5 only

dsyme:
OK that worked 👍

dsyme:
Note moving DiffSHarp.Core and the reference backend from `netstandard2.1` to `net5.0` was not necessary and probably not desireable - these should be kept independent of any particular runtime.

I'll likely revert that part of this change however it is farily harmless for now

dsyme:
Also, I believe we could have kept all the tests on netcoreapp3.1 and just used RollForward.

Issue #267, opened 2020-12-01
min/max API question


dsyme:
I have a question about max/min/argmax/argmin.

Today we have:

    a.argmax()  // returns index int[]
    a.max()  // returns scalar tensor
    a.max(b:Tensor)  // element-wise
    a.max(b:obj)  // "element-wise with data" via a.like(b)

1. This is missing an operation - "maximum with with a dimension to reduce" both for max and argmax

2. The last one looks wrong and the implicit invoke of `a.like(b)` is expensive. To be consistent with add etc. we would have `a.max(scalar)`

PyTorch has:

    a.max()
    a.max(dim)
    a.max(dim, keepdim=true)
    a.max(otherTensor)

Note that a scalar is *not* converted to a tensor implicitly here, because it would be ambiguous with `a.max(dim)`, a problem we will also have if we add `a.max(dim)`

The ambiguity is a potential cause for subtle DiffSharp bugs both in existing DiffSharp code and pytorch/numpy code brought across, where `a.max(0)` has completely different interpretations (usually caught by different return type).  

Because of this I propose:

    a.max()              // entire
    a.max(b:Tensor)  // element-wise
    a.maxScalar(b:scalar)   // element-wise with scalar
    a.max(dim:int, keepdim: bool)  // dimension-reducing - `dim` and `keepdim` must be explicit - return int[], Tensor pair

    a.argmax()   // return int[]
    a.argmax(dim:int, keepdim: bool)  // dimension-reducing - `dim` and `keepdim` must be explicit - return tensor

and same for min/argmin. This gives no ambguity. That would mean

* you can't write `a.max(0)`
* `a.max(0)` in existing DiffSharp code would have to be written `a.maxScalar(0)`
* `a.max(0)` in existing pytorch code would have to be written `a.max(0, keepdim=false)`

dsyme:
Separately I notice that this formula for min/max fails for unsigned byte tensors:

    member a.max(b:Tensor) = ((a + b) + Tensor.Abs(b - a)) / 2

    member a.min(b:Tensor) = ((a + b) - Tensor.Abs(a - b)) / 2

It is possible we should just build this in as a primitive in any case (since LibTorch exposes it as a primitive).

dsyme:
Closing as covered by #268

Issue #232, opened 2020-11-05
max/min along dimensions


dsyme:
We need versions of max/min which have a dimension/keepdim argument:

    torch.max(input, dim, keepdim=False, *, out=None) -> (Tensor, LongTensor)

dsyme:
Tracking this as one among a list of things in #227

PR #268, opened 2020-12-03
feature/argmaxdim - argmin/argmax along dimensions, adjust max/min of scalar

dsyme:
1. This addresses #232 for argmin/argmax along dimensions. These don't require differentiability as they only return tensors containing indexes, so are simpler than max/min along dimensions.

2. addresses part of #267 to adjust min/max in scalars

3. Fix a bug in max/min on integer tensors, where floating-point widening rules meant that float32 tensors were being produced.

4. Add testing for max/min and the new argmax/argmin along dimensions


gbaydin:
I checked the api changes here and the test cases carefully and I also did an independent check with pytorch to ensure things are fine. 

Everything looks good, except the comment https://github.com/DiffSharp/DiffSharp/issues/267#issuecomment-738039698 about min/max failing for unsigned byte tensors. Should we address this here in this PR as well?

Otherwise I'm ready to merge.

dsyme:
> Everything looks good, except the comment #267 (comment) about min/max failing for unsigned byte tensors. Should we address this here in this PR as well?

I consider this is a separate bug, I don't know the solution for it.  I'll open a separate issue on it.

dsyme:
Separate bug at https://github.com/DiffSharp/DiffSharp/issues/284

dsyme:
@gbaydin I'll leave you to merge

gbaydin:
Thank you, sounds good.

PR #273, opened 2021-02-13
DiffSharp.IO

gbaydin:
Introducing a new namespace called `DiffSharp.IO` to hold things like saving tensors to images, loading images as tensors, plotting, etc. Code importing and exporting sound and video files can go here as well. 

This also moves the dependence `SixLabors.ImageSharp` from `DiffSharp.Core` to `DiffSharp.IO` which I believe is better design for the core part of the library.


gbaydin:
Merging this now, thinking that we can revisit it later. @dsyme please let me know if you think the `DiffSharp.IO` design is the best we could do. The main motivation was to remove the `SixLabors.ImageSharp` dependency from `DiffSharp.Core`, but I think this IO project will also hold other pieces of code to get data in and out of DiffSharp down the line.

gbaydin:
Update: I decided to not follow this route, after I realized some cyclic dependencies this would introduce in some DiffSharp.Core code involving training models with image datasets. So this DiffSharp.IO project will subsequently disappear from dev.

dsyme:
As discussed in our call today, I think DiffSharp.IO or DiffSharp.Tools or DiffSharp.Utilities or DiffSHarp.Data makes a lot of sense for utility functionality with a lot of dependencies. The abstractions in Data.fs should go in DiffSharp.Core 

I do think you should do this one split - it's easy enough to manage.  I don't think you should split further until there is some other compelling need.

gbaydin:
Implemented in #280

PR #275, opened 2021-02-21
feature/nn4

gbaydin:
This is to keep track of more NN-related implementations.


gbaydin:
This part made significant improvements in convolution and transposed implementations, datasets, and nn module api. Merging now to dev as a preliminary step for project restructuring. Work will continue in another PR.

Issue #277, opened 2021-02-23
Need better implementation of tensor slicing via TorchSharp


gbaydin:
While working on other things and profiling cpu usage, I noticed that the `GetSlice` implementation in the Torch backend is somewhat unnecessarily complicated and it ends up consuming a lot of time (no actual figures to share at this point, sorry).

We keep doing a series of `Arange`, `IndexSelect`, and potentially `Squeeze` operations on an underlying Torch tensor.
https://github.com/DiffSharp/DiffSharp/blob/b8b6d3495a994309d76fbc3747e2684366d51e77/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs#L138

I believe that this should instead be implemented via a single call to the libtorch slicing API `torch::Tensor::index` which closely mirrors Python slicing format and https://pytorch.org/cppdocs/notes/tensor_indexing.html

I think the argument to `torch::Tensor::index` can be derived simply from the incoming `fullBounds:int[,]` argument to `GetSlice` 

Slicing is pretty critical and used a lot within the library. @dsyme I think doing this would require an update in TorchSharp as I couldn't locate the corresponding `torch::Tensor::index` wrapper there. There seems to be only `index_select` here https://github.com/xamarin/TorchSharp/blob/f456b0d1c162b6b7b4e7947df61fff3fea984d7d/src/Native/LibTorchSharp/THSTensor.cpp#L526

dsyme:
Yes please add a tracking issue in TorchSharp

gbaydin:
> Yes please add a tracking issue in TorchSharp

Ok done

gbaydin:
@dsyme following your comment here https://github.com/DiffSharp/DiffSharp/pull/368#discussion_r689131630 I was about to start working on reimplementing tensor slicing in the torch backend. 

But it looks like this was already implemented by you in this commit https://github.com/DiffSharp/DiffSharp/commit/93c8f38ac52ee5fff34d5f1231b3ce178399b8bb using `Tensor.index` https://github.com/xamarin/TorchSharp/blob/master/src/TorchSharp/Tensor/Tensor.cs#L1065

So I think we're good to go and can close this issue if I'm not missing something. Am I missing something?

gbaydin:
I'm convinced that this is already solved in existing code, so I'm closing the issue. :)

PR #278, opened 2021-02-23
Improvement in Torch backend slicing

gbaydin:
This relates to #277 and a performance issue with the way Torch backend slicing is currently implemented.

In the meantime, I have measured the modification in this PR to give approx. 2x speed up for a GAN prototype code I'm working on (which is mainly convolutions, transposed convolutions and lots of slicing within the reverse mode of these).


gbaydin:
Merging this as all tests pass and I have observed a significant performance benefit.

PR #281, opened 2021-03-01
feature/nn5

gbaydin:
This is to keep track of more ongoing ML work


gbaydin:
Merging this now as it includes the latest gan-related updates relevant for #285. Work will continue in a new PR.

PR #291, opened 2021-03-11
Use netstandard2.1 for the core

dsyme:
The core components on DiffSharp shouldn't depend on a specific .NET Runtime, they can just be netstandard2.1

This doesn't make any immediate difference but it's the right thing to do to prevent runtime-specific hacks creeping into the core abstractions 

Examples where we might want this

- we want to cross-compile to Javascript 
- we might want to host inside netstandard2.1 analyzers


gbaydin:
I've been reading these two pages to refresh my understanding on this. Sharing here for reference:

https://docs.microsoft.com/en-us/dotnet/standard/frameworks 
https://docs.microsoft.com/en-us/dotnet/standard/net-standard

PR #300, opened 2021-03-23
some docs work

dsyme:
@gbaydin For an internal presentation I need the DiffSharp site to have a little more content corresponding to its current capabilities.

We will need to iterate and revamp completely, but this is roughly what I need.

Let's discuss tomorrow.


gbaydin:
@dsyme sounds good, thank you!

PR #305, opened 2021-03-24
updates for models as functions

### LINKED ISSUE #292: Bug in Adam state step
This line is a bug, it is incrementing stateStep every time the updateRule is called, for each parameter.

https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Core/Optim.fs#L84

It should only be incremented once per iteration, not once per each parameter update.

This is important because we exponentiate by stateStep here: https://github.com/DiffSharp/DiffSharp/blob/dev/src/DiffSharp.Core/Optim.fs#L89

dsyme:
As discussed with @gbaydin 

Replaces `forwardCompose`, `forwardParameters`, `forwardLoss` with `model.asFunction` which takes parameters and input.

Also fixes #292 which requires optimizers having an updateStart entry for optimizers (I don't think there is any other choice, unless a per-parameter updateRule is removed

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "fsharp.formatting.commandtool": {
-      "version": "8.0.0",
+      "version": "9.0.4",
       "commands": [
         "fsdocs"
       ]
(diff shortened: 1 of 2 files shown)


dsyme:
@gbaydin The fix to the Adam optimizer caused a test failure

```
  Error Message:
     Expected: less than 22.705717086791992d
  But was:  28.78947639465332d

  Stack Trace:
     at Tests.TestModel.TestModelVAE() in /home/runner/work/DiffSharp/DiffSharp/tests/DiffSharp.Tests/TestModel.fs:line 914

```

I'll separate it out and we can discsuss.

PR #307, opened 2021-03-24
updates for models as functions

dsyme:
As discussed with @gbaydin

Replaces forwardCompose, forwardParameters, forwardLoss with model.asFunction which takes parameters and input.

### DIFF
--- a/src/DiffSharp.Core/Model.fs
+++ b/src/DiffSharp.Core/Model.fs
@@ -89,7 +89,10 @@ type ParameterDict() =
     ///  After this call the current parameters of the model will have attached derivatives for forward mode propagation.
     /// </remarks>
     member d.forwardDiff(derivatives:ParameterDict, ?tag:uint32) = 
-        d.iter(fun (n, p) -> p.forwardDiff(derivatives.[n], ?tag=tag))
+        // This is to be extra cautious about all Parameters in the ParameterDict getting the same tag, which is crucial for correctness of differentiation results
+        // If we leave the default tag value to be determined by each underlying tensor, there is a risk that the tag can somehow change during the ParameterDict .iter call
+        let tag = defaultArg tag GlobalNestingLevel.Current
+        d.iter(fun (n, p) -> p.forwardDiff(derivatives.[n], tag=tag))
 
     /// <summary>
     ///  Adjust the parameters to include support for reverse-mode automatic differentiation.
@@ -101,7 +104,10 @@ type ParameterDict() =
     ///  will be available. 
     /// </remarks>
     member d.reverseDiff(?tag:uint32) = 
-        d.iter(fun (_, p) -> p.reverseDiff(?tag=tag))
+        // This is to be extra cautious about all Parameters in the ParameterDict getting the same tag, which is crucial for correctness of differentiation results
+        // If we leave the default tag value to be determined by each underlying tensor, there is a risk that the tag can somehow change during the ParameterDict .iter call
+        let tag = defaultArg tag GlobalNestingLevel.Current
+        d.iter(fun (_, p) -> p.reverseDiff(tag=tag))
 
     /// <summary>TBD</summary>
     member d.noDiff() = d.iter(fun (_, p) -> p.noDiff())
--- a/src/DiffSharp.Core/Model.fs
+++ b/src/DiffSharp.Core/Model.fs
@@ -80,15 +80,28 @@ type ParameterDict() =
     /// <summary>TBD</summary>
     member d.iter(f:string*Parameter->unit) = for KeyValue(n, p) in d.values do f(n,p)
 
-    /// <summary>TBD</summary>
+    /// <summary>
+    ///  Adjust the parameters to include support for forward-mode automatic differentiation.
+    /// </summary>
+    /// <param name="derivatives">The derivatives of the parameters</param>
+    /// <param name="tag">The level tag for nested differentiation.  Defaults to the current global nesting level</param>
+    /// <remarks>
+    ///  After this call the current parameters of the model will have attached derivatives for forward mode propagation.
+    /// </remarks>
     member d.forwardDiff(derivatives:ParameterDict, ?tag:uint32) = 
-        let tag = defaultArg tag GlobalNestingLevel.Current
-        d.iter(fun (n, p) -> p.forwardDiff(derivatives.[n], tag))
+        d.iter(fun (n, p) -> p.forwardDiff(derivatives.[n], ?tag=tag))
 
-    /// <summary>TBD</summary>
+    /// <summary>
+    ///  Adjust the parameters to include support for reverse-mode automatic differentiation.
+    /// </summary>
+    /// <param name="tag">The level tag for nested differentiation.  Defaults to the current global nesting level</param>
+    /// <remarks>
+    ///  After this call the current parameters of the model will support reverse-mode propagation. After the completion
+    ///  of the corresponding <c>reverse</c> operation, the computed derivative
+    ///  will be available. 
+    /// </remarks>
     member d.reverseDiff(?tag:uint32) = 
-        let tag = defaultArg tag GlobalNestingLevel.Current
-        d.iter(fun (_, p) -> p.reverseDiff(tag))
+        d.iter(fun (_, p) -> p.reverseDiff(?tag=tag))
 
     /// <summary>TBD</summary>
     member d.noDiff() = d.iter(fun (_, p) -> p.noDiff())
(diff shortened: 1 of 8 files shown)


gbaydin:
Awesome, thanks for putting this together.

src/DiffSharp.Core/Model.fs
```diff
@@ -80,15 +80,28 @@ type ParameterDict() =
...
+    /// <summary>
+    ///  Adjust the parameters to include support for forward-mode automatic differentiation.
+    /// </summary>
+    /// <param name="derivatives">The derivatives of the parameters</param>
+    /// <param name="tag">The level tag for nested differentiation.  Defaults to the current global nesting level</param>
+    /// <remarks>
+    ///  After this call the current parameters of the model will have attached derivatives for forward mode propagation.
+    /// </remarks>
     member d.forwardDiff(derivatives:ParameterDict, ?tag:uint32) = 
-        let tag = defaultArg tag GlobalNestingLevel.Current
```
dsyme, line 85: Note the default is applied in `p.forwardDiff`

src/DiffSharp.Core/Model.fs
```diff
@@ -80,15 +80,28 @@ type ParameterDict() =
...
+    ///  Adjust the parameters to include support for reverse-mode automatic differentiation.
+    /// </summary>
+    /// <param name="tag">The level tag for nested differentiation.  Defaults to the current global nesting level</param>
+    /// <remarks>
+    ///  After this call the current parameters of the model will support reverse-mode propagation. After the completion
+    ///  of the corresponding <c>reverse</c> operation, the computed derivative
+    ///  will be available. 
+    /// </remarks>
     member d.reverseDiff(?tag:uint32) = 
-        let tag = defaultArg tag GlobalNestingLevel.Current
```
dsyme, line 90: Note the default is applied in p.reverseDiff

dsyme:
@gbaydin This is ready - not the fix to the docs build as well

src/DiffSharp.Core/Model.fs
gbaydin: I think the reason for this line was to ensure all tensors within the ParameterDict will get the same tag, which is absolutely crucial. I think your change still preserves that in a single-threaded setting but it feels less safe.

Would you mind if I revert this with a comment to explain the reason?

src/DiffSharp.Core/Model.fs
gbaydin: Please see the comment I left for the `forwardDiff` case.

gbaydin:
I had a more careful look at the naming. I now remember the reason for calling this `.forwardParameters` and I think we shouldn't use the `.asFunction` name.

The reason is that `.forward` and `.asFunction` are very related, they are two sides of the same coin or two aspects of the same underlying function. I think the name should reflect the connection and `.forward` and `.asFunction` names don't imply any sort of connection.

More precisely, a NN model is the function `y = f(x, theta)` where y` is the output, `x` is the input, `theta` are the parameters. 

* `.forward` is `y = f(x, theta*)` the output as a function of the input for a fixed `theta*`
* `.forwardParameters` or `.asFunction` is the whole `y = f(x, theta)`.
* there is also the third one which we don't expose: `y = f(x*, theta)` the output as a function of the parameters for a fixed input. 

Ideally we should find two names (or three names if we expose the third) for these cases that give the user the feeling that they are related. `.forward` is pretty fixed and we shouldn't change it. So this was the reason to call it `.forwardParameters`.

If we name all three above, `.forwardParameters` is a better name for the third one. The second one, that you called `.asFunction`, could be named `.forwardFull`, `.forwardAll` or something similar.

dsyme:
Note this is only a priority if it forms part of the way we write models.

I think most of the problem my cognitive issues with the names stems with the use of the name 'forward'  which is already just a bit obscure because of multiple meanings of "forward" and this kind of compounds that.  Here's what I think when I simply read the words plainly:

* To me "forward" is explained by PyTorch's "forward pass before we do a backward pass"  or "run the forward computation" or "forward the inputs"

* To me "forwardParameters" sounds like it will "forward" some the parameters or possibly it's the "forwards-direction parameters" as opposed to come other kind of parameters.  It doesn't sound like something that will adjust mutable state hidden in the model

* To me "forwardCompose" sounds like it will compose something in the forward direction as opposed to composing something in the backward direction (forward as adjective - and is it sounds weird)

* To me "forwardLoss" sounds like it computes the loss incurred running the forward (which is half correct but the you need to provide the loss function to "forwardLoss" and that's really strange)

* None of the names indicate anything like what's actually happening, compared to "asFunction" or "toFunction" 

* Anything that implicitly sets/gets the mutable parameters state hidden in each model object should be clearly identifiable,  preferably via one searchable thing.

Anyway I don't mind too much as long as these aren't common things to do. These names seem reasonable to me:

* m.forward x = ... 

* m.asFunction x p = ...   

* m.asFunctionOfInput = let p = m.parametersVector in (fun x -> m.asFunction x p)

Note `m.asFunctionOfParameters` is just the partial application of `m.asFunction` and probably doesn't deserve a separate method.

gbaydin:
I think in your last list with three items `m.asFunctionOfInput` should be the same thing with `m.forward`. In other words `m.forward` is a function of the input (taking fixed parameters as fixed).

src/DiffSharp.Core/Model.fs
dsyme: Yes that's fine

dsyme:
> I think in your last list with three items m.asFunctionOfInput should be the same thing with m.forward. In other words m.forward is a function of the input (taking fixed parameters as fixed).

Right, in the absence of tensor or parameter-dictionary mutation (including by calling `asFunction`) then `m.asFunctionOfInput`  = `m.forward`.   

The problem of course is that parameter dictionary mutation is going to heppen. A lot! Whether by calling `asFunction` or optimizing the model or setting the parameters explicitly.  In this case taking a snapshot of the parameters before returning the closure gives a more correct mathematical function, though slower.  

TBH this is all just bandaids over the way we author models w.r.t. a single mutable parameter dictionary.  This is going to be a recurring problem I think.  It would be kind of great if there were some way to recover the mathematical functions and be able to free ourselves from the potential of interacting with that mutation.  One technique is to support a deep-clone down the model structure, e.g.

    m.forward x = ...

    m.asFunction x p = m.clone(p).forward(x)

    m.asFunctionOfInput = let p = m.parametersVector in (fun x -> m.asFunction x p)

The difference here is that `asFunction` is mutating a different, private set of model objects and parameter dictionary.  This gives the true function.

There are likely also other ways to do this sort of thing - this gets into self-adjusting-code and copy-on-write territory, i.e. preseving cheap functional snapshots of the parameter dictionary while still allowing subsequent mutations.  But those techniques complicate the programming model considerably.

gbaydin:
I agree with most of the points, but I don't see what you mean (or why you introduce) the `m.asFunctionOfInput` in your points. By definition `m.asFunctionOfInput` is `m.forward` and these things are identical. `m.forward` is the the function of input with parameters kept fixed.

dsyme:
> By definition m.asFunctionOfInput is m.forward and these things are identical. m.forward is the the function of input with parameters kept fixed.

"forward" depends on the model parameter dictionary mutable state, but [this version of "asFunctionOfInput"](https://github.com/DiffSharp/DiffSharp/pull/307#issuecomment-808781893) doesn't.   Compare for example:

    let f = m.forward
    let v1 = f(someTensor)
    m.parametersVector <- someOtherTensor
    let v2 = f(someTensor) // changes

v.

    let f = m.asFunctionOfInput() // that is, "asFunctionOfInputWithFixedParameters"
    let v1 = f (someTensor)
    m.parametersVector <- someOtherTensor
    let v2 = f(someTensor) // same as v1

For the first, v1 and v2 is different. For the second, they are the same, as the parameters are captured (fixed) at the point `asFunctionOfInput` is called.  That makes it a true mathematical function independent of state accessible for mutation.  

You can also achieve this by replacing the clone with entirely new model instances:

    let m2 = new Model()
    m2.parametersVector <- m.parametersVector
    let f = m2.forward
    let v1 = f (someTensor)
    m.parametersVector <- someOtherTensor
    let v2 = f(someTensor) // same as v1, independent of mutation

As an aside, any use of in-place tensor mutation obviously makes this stuff harder to reason about, it's the reason I think we should strongly limit the places where we allow tensor mutation, even if it means more copy-in-copy-out in some places.

gbaydin:
Oh I see it now, it took me a while to focus on the actual implementation you were talking about. I was too focused on the abstract mathematical definitions I had in the beginning. Thank you!

gbaydin:
I'm going to merge this very soon with the `.asFunction` name. Looking at a few final things now.

PR #311, opened 2021-03-31
Simpler extension API

gbaydin:
This is work in progress to deliver a much simplified extension api. Based on the prototype in #89 which served as a great pathfinder for the design. I'm really happy with the simplicity of the elementwise case. Still a few things need to be added and tested.

```fsharp
type Tensor with
    member a.sin() = 
        Tensor.Op
            { new UnaryOpElementwise() with 
                member _.fRaw(a) = a.SinT()
                member _.df_da(a,f) = a.cos()
            }
            (a)

    member a.cos() =
        Tensor.Op
            { new UnaryOpElementwise() with
                member _.fRaw(a) = a.CosT()
                member _.df_da(a,f) = -a.sin()
            }
            (a)

    member a.exp() =
        Tensor.Op
            { new UnaryOpElementwise() with
                member _.fRaw(a) = a.ExpT()
                member _.df_da(a,f) = f
            }
            (a)

    member a.log() =
        Tensor.Op
            { new UnaryOpElementwise() with
                member _.fRaw(a) = a.LogT()
                member _.df_da(a,f) = 1/a
            }
            (a)

    member a.transpose() =
        Tensor.Op
            { new UnaryOp() with
                member _.fRaw(a) = a.TransposeT2()
                member _.ad_df_da(a,ad,f) = ad.transpose()
                member _.fd_df_da(a,f,fd) = fd.transpose()
            }
            (a)

    member a.pow(b) =
        Tensor.Op
            { new BinaryOpElementwise() with
                member _.fRaw(a,b) = a.PowTT(b)
                member _.df_da(a,b,f) = b * f / a
                member _.df_db(a,b,f) = f * a.log()
            }
            (a,b)

    member a.mul(b) =
        Tensor.Op
            { new BinaryOpElementwise() with
                member _.fRaw(a,b) = a.MulTT(b)
                member _.df_da(a,b,f) = b
                member _.df_db(a,b,f) = a
            }
            (a,b)

    member a.matmul(b) =
        Tensor.Op
            { new BinaryOp() with
                member _.fRaw(a,b) = a.MatMulTT(b)
                member _.ad_df_da(a,ad,b,f) = ad.matmul(b)
                member _.bd_df_db(a,b,bd,f) = a.matmul(bd)
                member _.fd_df_da(a,b,f,fd) = fd.matmul(b.transpose())
                member _.fd_df_db(a,b,f,fd) = a.transpose().matmul(fd)
            }
            (a,b)
```


dsyme:
Great to see this, it looks fine to me

dsyme (approved):

src/DiffSharp.Core/Tensor.fs
```diff
@@ -2908,3 +2916,63 @@ and TensorOp =
     | AcosT of Tensor
     | AtanT of Tensor
     | NewT
+    | OpUnaryT of Tensor*(Tensor*Tensor*Tensor->Tensor)
+    | OpBinaryTT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor*Tensor)
+    | OpBinaryTC of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
+    | OpBinaryCT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
+
+[<AbstractClass>]
+type UnaryOp() =
```
dsyme, line 2925: Add comments and example

src/DiffSharp.Core/Tensor.fs
```diff
@@ -2908,3 +2916,63 @@ and TensorOp =
...
+    | OpUnaryT of Tensor*(Tensor*Tensor*Tensor->Tensor)
+    | OpBinaryTT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor*Tensor)
+    | OpBinaryTC of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
+    | OpBinaryCT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
+
+[<AbstractClass>]
+type UnaryOp() =
+    abstract fRaw: a:RawTensor->RawTensor
+    abstract ad_df_da: a:Tensor*ad:Tensor*f:Tensor->Tensor
+    abstract fd_df_da: a:Tensor*f:Tensor*fd:Tensor->Tensor
```
dsyme, line 2928: Add comments to each of these

src/DiffSharp.Core/Tensor.fs
```diff
@@ -2908,3 +2916,63 @@ and TensorOp =
...
+    abstract df_da: a:Tensor*b:Tensor*f:Tensor->Tensor
+    abstract df_db: a:Tensor*b:Tensor*f:Tensor->Tensor
+    override op.ad_df_da(a,ad,b,f) = ad*op.df_da(a,b,f)
+    override op.bd_df_db(a,b,bd,f) = bd*op.df_db(a,b,f)
+    override op.fd_df_da(a,b,f,fd) = fd*op.df_da(a,b,f)
+    override op.fd_df_db(a,b,f,fd) = fd*op.df_db(a,b,f)
+
+
+type Tensor with
+    static member Op(ext: UnaryOp) =
```
dsyme, line 2960: Add `///` comments

dsyme:
Note I will replay #252 over this once it's in.  Really glad to have this addressed, it's so crucial in the long term

gbaydin:
@dsyme one thing I need advice with is this:

Currently the list of ops in https://github.com/DiffSharp/DiffSharp/blob/4c8df05a9ceb78c8f76d787117ba19db4f1ebad0/src/DiffSharp.Core/Tensor.fs#L2807

is serving an important purpose of keeping track of operation history in the computation graph of reverse mode. This is immensely useful and needed for debugging. It's analogous to the PyTorch case below, for example.

```python
import torch
x = torch.tensor([1.,2,3], requires_grad=True)
y =x.sum()
print(y) 
```
```
tensor(6., grad_fn=<SumBackward0>)
```

When we introduce the extension api, we lose this and all operations will rely on the generic `OpUnary`, `OpBinary` etc cases. We need to introduce some structure to keep track unique operation type labels for each extension operation. Do you have a suggestion?

I'm thinking about adding a string to all cases under `TensorOp`, but this feels like a bad idea. I think whatever we introduce should apply to existing ops in `TensorOp` as well as ops that are implemented through the new extension api.

dsyme:
> I'm thinking about adding a string to all cases under TensorOp, but this feels like a bad idea. I think whatever we introduce should apply to existing ops in TensorOp as well as ops that will be implemented through the new extension api.

Yes I understand. TBH I think it's best just to add a string to these cases:

```
    | OpUnaryT of Tensor*(Tensor*Tensor*Tensor->Tensor)
    | OpBinaryTT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor*Tensor)
    | OpBinaryTC of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
    | OpBinaryCT of Tensor*Tensor*(Tensor*Tensor*Tensor*Tensor->Tensor)
```

and implement a "ToString()" on the type which picks up either that string (for those cases) or the `GetType().Name` (for the other cases)

If we move various things out of Tensor.fs into extensions (e.g. move the maxpool out, like avgpool) then we can add the strings at that point.

gbaydin:
This is ready to go. I'm quite happy about how simple and nice this turned out! I think it will be really useful for ongoing work.

I'm merging this now thinking that there are other PRs need to be updated to incorporate this.

I added some basic documentation here and I will handle the complete documentation including a tutorial about how to create new ops as the next PR.

PR #315, opened 2021-04-13
Update fsdocs

dsyme:
Updates fsdocs to latest which allows API cross-references to be used

This update also adds buttons to copy the markdown or XML text needed for cross references, e.g. 

image

We should likely use a "link" logo rather than a clipboard but hey

These are currently added to every member and entiry in the generated lists.  This makes them a little visually intrusive but we should just live with it for now.  When we switch to generating a page for each member we can switch to having these on each actual target page rather than in the long generated lists

### DIFF
--- a/.config/dotnet-tools.json
+++ b/.config/dotnet-tools.json
@@ -3,7 +3,7 @@
   "isRoot": true,
   "tools": {
     "fsharp.formatting.commandtool": {
-      "version": "10.0.11",
+      "version": "10.0.17",
       "commands": [
         "fsdocs"
       ]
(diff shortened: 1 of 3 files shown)


dsyme:
There are a few other updates like generating "dt" elements for the metadata at the top of each page for each type/module.

I'll merge this and we can see how the docs look

dsyme:
Hmm this caused doc generation to fail.  I'll revert and explore

```
  generating model for docs/getting-started-libtorch.fsx --> ./getting-started-libtorch.html
Error: Process completed with exit code 1.
```

PR #331, opened 2021-04-22
Main readme deadlink fix 

kevmal:
fix torch ipynb/fsx link

PR #332, opened 2021-04-22
Make PR github actions run in the target branch

gbaydin:
This intends to fix the failing CI code coverage upload exemplified in #331. At first I was planning to make secrets available to PRs using the new `pull_request_target` event: https://nathandavison.com/blog/github-actions-and-the-threat-of-malicious-pull-requests 

But later I changed my mind and now think that we can just disable code coverage for PRs altogether. This was prompted by reading https://securitylab.github.com/research/github-actions-preventing-pwn-requests/

So for the time being let's keep this safe and simple without any possibility of repo secrets leaking to forker repo builds.

### DIFF
--- a/.github/workflows/pull-request-debug.yml
+++ b/.github/workflows/pull-request-debug.yml
@@ -1,7 +1,7 @@
 name: Build/test Debug
 
 on:
-  pull_request_target:
+  pull_request:
     branches:
       - '**'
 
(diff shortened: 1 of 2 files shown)


gbaydin:
@dsyme I'm merging this. Let me know what you think and we can revert it as well.

dsyme:
Sounds right to me :-)

PR #338, opened 2021-06-06
Documentation 3

gbaydin:
Continuing documentation work


docs/index.fsx
```diff
@@ -1,42 +1,61 @@
...
 (**
+[![Binder](https://colab.research.google.com/assets/colab-badge.svg)](https://colab.research.google.com/github/DiffSharp/diffsharp.github.io/blob/master/{{fsdocs-source-basename}}.ipynb)&emsp;
 [![Binder](img/badge-binder.svg)](https://mybinder.org/v2/gh/diffsharp/diffsharp.github.io/master?filepath={{fsdocs-source-basename}}.ipynb)&emsp;
 [![Script](img/badge-script.svg)]({{fsdocs-source-basename}}.fsx)&emsp;
 [![Script](img/badge-notebook.svg)]({{fsdocs-source-basename}}.ipynb)
 
 # DiffSharp: Differentiable Tensor Programming Made Simple
 
-DiffSharp is a tensor library with support for [differentiable programming](https://en.wikipedia.org/wiki/Automatic_differentiation).
+DiffSharp is a tensor library with support for [differentiable programming](https://en.wikipedia.org/wiki/Differentiable_programming).
```
dsyme, line 41: That's a much better link to use

docs/index.fsx
```diff
@@ -1,42 +1,61 @@
...
+<button class="button" style="vertical-align:middle" onclick="window.location.href='{{root}}install.html'"><span>Install »</span></button>
+
+## Key Features
+
 🗹 Nested and mixed-mode differentiation
 
 🗹 Common optimizers, model elements, differentiable probability distributions
 
-🗹 F# for robust functional programming 
+🗹 F# programming language for robust functional programming
```
dsyme, line 52: The repeated 'programming' doesn't look right, I think the former is better?

docs/index.fsx
```diff
@@ -47,66 +66,69 @@ DiffSharp provides world-leading automatic differentiation capabilities for tens
...
 
 t3.conv2d(t4)
+(*** include-it ***)
 
 (** 
 Take the gradient of a vector-to-scalar function:
 *)
 
-let f (x: Tensor) = x.exp().sum()
+let f (x:Tensor) = x.exp().sum()
```
dsyme, line 118: Recommend space after `:` as standard formatting

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+
+// Inspect data as ASCII and show label
+printfn "Data: %A\nLabel: %A" (data.toImageString()) label
+(*** include-output ***)
+
+(**
+ 
+### Data Loaders
+
+A data loader handles tasks such as constructing minibatches from an underlying dataset on-the-fly, shuffling the data, and moving the data tensors between devices. In the example below we show a single batch of six MNIST images and their corresponding classification labels.
```
dsyme, line 86: More information on devices and data loaders would be useful, e.g. guidance on how many tensor copies to expect when getting data to GPU etc.

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+
+let loader = DataLoader(dataset, shuffle=true, batchSize=6)
+let batch, labels = loader.batch()
+
+printfn "%A\nLabels: %A" (batch.toImageString()) labels
+(*** include-output ***)
+
+(**
+
+In practice a data loader would be typically used to iterate over all minibatches in a given dataset in order to feed each minibatch through a machine learning model. One full iteration over the dataset would be called an "epoch". Typically you would perform multiple such epochs of iterations during the training of a model.
```
dsyme, line 98: "is typically used" rather then "would be typically used"

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+        // Process the minibatch
+        // ...
+(**
+
+## Models
+
+Many machine learning models are differentiable functions whose parameters can be tuned via [gradient-based optimization](https://en.wikipedia.org/wiki/Gradient_descent), finding an optimum for an objective function that quantifies the fit of the model to a given set of data. These models are typically built as compositions non-linear functions and ready-to-use building blocks such as linear, recurrent, and convolutional layers.
+
+DiffSharp provides the most commonly used model building blocks including convolutions, transposed convolutions, batch normalization, dropout, recurrent and other architectures.
+
```
dsyme, line 115: I think add a comment that in DiffSharp these parameters are tensors that may themselves be subject to differentiation and gradient-taking, rather than being floating-point parameters.

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+        |> dsharp.tanh
+
+// Define a model class inheriting the base
+type Discriminator(nz:int) =
+    inherit Model()
+    let fc1 = Linear(28*28, 1024)
+    let fc2 = Linear(1024, 512)
+    let fc3 = Linear(512, 256)
+    let fc4 = Linear(256, 1)
+    do base.add([fc1; fc2; fc3; fc4])
```
dsyme, line 154: Maybe put a comment on this line, since it's necessary

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+let dis = Discriminator(nz)
+
+print gen
+print dis
+(*** include-output ***)
+
+(**
+### Constructing models, DiffSharp style
+
+DiffSharp's real strength lies in the [functional programming](https://en.wikipedia.org/wiki/Functional_programming) paradigm enabled by the F# language, where functions are first-class citizens, many algorithms can be constructed by applying and composing functions, and differentiation operations can be expressed as composable [higher-order functions](https://en.wikipedia.org/wiki/Higher-order_function). This allows very succinct (and beautiful) machine learning code to be expressed as a powerful combination of [lambda calculus](https://en.wikipedia.org/wiki/Lambda_calculus) and [differential calculus](https://en.wikipedia.org/wiki/Differential_calculus).
```
dsyme, line 182: "real" seems a little odd here.  It's an important strength, but I'd rate the nested differentiation capablities as more important

docs/quickstart.fsx
```diff
@@ -0,0 +1,229 @@
...
+
+For example, the following constructs the same GAN architecture (that we constructed in PyTorch style in the previous section) using DiffSharp's `-->` differentiable composition operator, which allows you to seamlessly compose `Model` instances and differentiable `Tensor->Tensor` functions.
```
dsyme, line 184: I think "composition operator" is better than "differentiable composition operator".  There's nothing particularly differentiable about this more than other DiffSharp constructs?

dsyme:
It's looking great, I left some comments

docs/index.fsx
gbaydin: I agree, we can revert. I was thinking about people read "F#" for the first time in their lives.

docs/index.fsx
gbaydin: Personally I find `x:Tensor` to be much smoother (doesn't break the flow), but if we adopt `x: Tensor` as a style, I can adapt to that. It's a minor thing.

docs/quickstart.fsx
gbaydin: The idea is that the quickstart page is just about the basics and getting up and running. I think we can have the full "using devices" guidance in another page and link to it from here. Alternatively we can also cover GPU usage in quickstart.

docs/quickstart.fsx
gbaydin: The "parameters" in line 112 refers to tensors that are the parameters of neural networks. I think you're talking about the hyperparameters of the optimization algorithms, which we should talk about in the way you're suggesting. I think we do that elsewhere in this page.

docs/quickstart.fsx
gbaydin: I agree.

As a separate idea: can we improve the behavior in a way that this `base.add` is not needed? In Python, there is an automated way that the class discovers any attributes that are other NN modules and they automatically are registered as submodules. 

Discussed here:

https://discuss.pytorch.org/t/how-does-a-module-know-about-its-component-modules/13236

and implemented here as a custom `__setattr__`
https://github.com/pytorch/pytorch/blob/394dd391dd5c30de371e77722d5d0261dd90c0b0/torch/nn/modules/module.py#L1138

Can we have a similar thing in F#?

docs/quickstart.fsx
gbaydin: I'm trying to highlight that `-->` is for composing differentiable transformations (either `Tensor->Tensor` functions or `Module`s). If the things being composed are not differentiable, there isn't much we can do with that. I really want to highlight the "differentiable" bit, how can we say it in a better way?

gbaydin:
Thanks @dsyme ! I think I will address these comments and merge it today so that the quickstart and the improvements to the home page are public at least.

docs/quickstart.fsx
dsyme: Then I think add a separate sentence saying that `-->` maintains differentiability.

docs/quickstart.fsx
dsyme: Yes that's right

docs/quickstart.fsx
dsyme: Yeah it's hard to remove this without resorting to reflection hackery. I've thought about it a few times and haven't come up with anything

PR #340, opened 2021-06-21
bump versions

dsyme:
Note to self: The libtorch-cpu package is now split into

* libtorch-cpu-win-x64
* libtorch-cpu-linux-x64
* libtorch-cpu-macos-x64
* libtorch-cpu referencing the last three

DiffSharp-cpu references the last one, and is simple, but means the user getting started download all three of the other packages.   We could create corresponding bundles 

* DiffSharp-cpu-win-x64
* DiffSharp-cpu-linux-x64
* DiffSharp-cpu-macos-x64

### DIFF
--- a/.github/workflows/pull-request-debug.yml
+++ b/.github/workflows/pull-request-debug.yml
@@ -6,9 +6,9 @@ on:
       - '**'
 
 jobs:
-  build:
+  build_windows:
 
-    runs-on: ubuntu-latest
+    runs-on: windows-latest
 
     steps:
     - uses: actions/checkout@v2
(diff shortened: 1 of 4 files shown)


dsyme:
We will also need to check 

* [ ] OSX CPU 
* [ ] Windows CUDA 
* [ ] Linux CUDA

I'll add a CI build for the first.  Am trying the second on my local machine but download times are long

dsyme:
The OSX libtorch packages aren't working, I'll get them fixed

dsyme:
Oh cool, that now works

I'll merge this.

dsyme:
OK this passes GPU tests on Windows, I will pull this now.

dsyme:
So DiffSharp now has Mac support :)

Issue #336, opened 2021-05-26
Tensor.variance(dim:int) bug: sometimes it gives a negative variance


nhirschey:
There appears to be a bug in `Tensor.variance(dim:int, ?keepDim:bool, ?unbiased:bool)` such that it can give a negative variance.

```fsharp
#r "nuget: DiffSharp-cuda-windows, 1.0.0-preview-783523654"

open DiffSharp

dsharp.config(dtype=Dtype.Float32,device=Device.CPU,backend=Backend.Torch)

let test =
    dsharp.tensor([[996.52, -8.02],
        [996.57, -8.41],
        [996.53, -8.51],
        [996.51, -8.31],
        [996.51, -8.27],
        [996.5, -8.05],
        [996.5, -7.62],
        [996.5, -7.62]],dtype=Float32)

// Here we get a negative variance for the first column of data.
test.variance(0)
(** output:
val it : Tensor = tensor([-0.0714286, 0.115278])
*)

// Here we get correct variance
test.[0..,0].variance()
(** output:
val it : Tensor = tensor(0.000564476) {Item = ?;
                                       backend = Torch;
                                       depth = 0;
                                       derivative = ?;
                                       derivativeDeep = ?;
                                       device = Device (CPU, -1);
                                       deviceType = CPU;
                                       dim = 0;
                                       dtype = Float32;
                                       fanout = ?;
                                       nelement = 1;
                                       nestingTag = ?;
                                       parentOp = ?;
                                       primal = ...;
                                       primalDeep = ...;
                                       shape = [||];}
*)


```

And a sanity check:
```fsharp
let correctVariance =
    Array.init test.shape.[0] (fun i ->  test.[i,0].toSingle())
    |> fun xs ->
        let mu = xs |> Array.average
        let n = xs.Length - 1 |> single
        xs |> Array.sumBy(fun x -> (1f/n)*(x-mu)**2.0f)

(** output:
val correctVariance : float32 = 0.000564472517f
*)
```

dsyme:
Note sum/variance need to be reimplemented https://github.com/DiffSharp/DiffSharp/issues/279

gbaydin:
@nhirschey thank you for catching and isolating this!

I think this is happening because the `.variance()` and `.variance(dim:int)` versions used different implementations, referred to as "two-pass" and "naive" in this page https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance

[...]

PR #357, opened 2021-07-26
Fix numerical precision of the variance(dim) calculation

nhirschey:
The prior naive variance(dim) algorithm had issues with numerical precision. See https://github.com/DiffSharp/DiffSharp/issues/336

This uses Welford's online algorithm and includes tests to catch numerical precision issues. I also added a test case to catch precision issues when the algorithm is rewritten (the precision test is only done for Float64).

The algorithm is discussed on [wikipedia](https://en.wikipedia.org/wiki/Algorithms_for_calculating_variance#Welford's_online_algorithm) with the original source [here](https://www.jstor.org/stable/1266577).

This is the same algorithm used by FSharp.Stats ([source link](https://github.com/fslaborg/FSharp.Stats/blob/a4005abc1232e5336c31ee6dae0e6179eb848407/src/FSharp.Stats/Seq.fs#L316))


gbaydin:
@nhirschey this is looking great! 

I checked the fix and the test case you supplied. I also confirmed the correct behavior with PyTorch.

Issue #93, opened 2020-04-13
Implement norm and matrix inverse


gbaydin:
Can be equivalent to https://pytorch.org/docs/stable/torch.html#torch.norm

cgravill:
https://pytorch.org/docs/stable/linalg.html#torch.linalg.norm would be useful for some code I'm trying to port to DiffSharp 1.0

gbaydin:
Matrix inverse implemented in #364

gbaydin:
TorchSharp updated in #368 to latest to introduce the norms needed to close this issue.

gbaydin:
Norm implemented in #369.

PR #364, opened 2021-08-02
Matrix inverse, solve, outer product

gbaydin:
Work in progress for matrix inverse #93


gbaydin:
This implemented matrix inverse. In the process I also had to implement solve (solution of linear equations), outer product, and a minor update o unsqueeze (#366).

PR #368, opened 2021-08-15
Update torchsharp to 0.91.52719

gbaydin:
This is updating TorchSharp to 0.91.52719 which has lots of breaking changes due to mass renamings: https://github.com/xamarin/TorchSharp/pull/302

This is needed going forward with latest TorchSharp, needed for example for `torch.linalg.matrix_norm` and `torch.linalg.vector_norm` needed for DiffSharp 1.0.

### DIFF
--- a/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
+++ b/src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
@@ -10,7 +10,6 @@ open DiffSharp
 open DiffSharp.Backends
 open DiffSharp.Util
 open TorchSharp
-// open TorchSharp.Tensor
 
 type TorchShape = int64[]
 type TorchDevice = Torch.Device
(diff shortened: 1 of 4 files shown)


src/DiffSharp.Backends.Torch/Torch.RawTensor.fs
```diff
@@ -124,7 +123,6 @@ type TorchRawTensor(tt: TorchTensor, shape: Shape, dtype: Dtype, device: Device)
 
     override t.GetSlice(fullBounds:int[,]) =
         let n = fullBounds.GetLength(0)
```
dsyme, line 125: We can update this to use tensor slicing now https://github.com/xamarin/TorchSharp/blob/master/src/TorchSharp/Tensor/Tensor.cs#L1065

For setting slices: https://github.com/xamarin/TorchSharp/blob/master/src/TorchSharp/Tensor/Tensor.cs#L1095

gbaydin: Cool, thanks for reminding. We can do this now and close #277

PR #369, opened 2021-08-19
Vector norm

gbaydin:
Implementing vector and matrix norm as needed in #93


gbaydin:
Vector norm is working and matrix form is left for later as it's not used a lot in practice.

Issue #239, opened 2020-11-11
div operator v. floor_div in TorchLib 1.7.0


dsyme:
Upgrading to TorchLib 1.7.0 indicates a change in PyTorch, 

Previously `/` on integer tensors was integer division (floor_div).   Now it promotes to float32

```python
>>> (torch.tensor([1],dtype=torch.int8) / torch.tensor([1],dtype=torch.int8)).dtype
torch.float32
```

We should likely adjust to match PyTorch.  For now I'll adjust to cast back to the integer type

PR #373, opened 2021-08-28
Implement "true division" promoting integers to floating point

gbaydin:
Addresses #239 and introduces behavior consistent with PyTorch > 1.7.

In PyTorch this is called "true division": https://pytorch.org/docs/stable/generated/torch.div.html

The behavior is ultimately due to: https://www.python.org/dev/peps/pep-0238/


gbaydin:
While implementing this, I noticed a bug in TorchSharp `Scalar` creation where the types of numeric values were not preserved (e.g., creating a `Scalar` from an `int8` gives us `int64`). The issue is reported here: https://github.com/xamarin/TorchSharp/issues/331

Issue #204, opened 2020-10-01
Run samples cleanly on GPU with one line addition


dsyme:
@gbaydin I'm just wondering what your expectation/desire is here for the programming model

I'm looking at the VAE.fsx sample and trying to run it on GPU. Is the intent that I do this by setting GPU as the default device?

    dsharp.config(backend=Backend.Torch, device=Device.GPU)

or is it expected to instruct things via `model.move(Device.GPU)`?  In which case how do we specify the move of the data and related tensors to the GPU? 

When I make the GPU the default then  "Saving samples" takes a very long time. Adding this helps:

    samples.move(Device.CPU).saveImage(sprintf "samples_%A_%A.png" epoch i)

We should probably always move to the CPU before doing things like saveImage

dsyme:
As an aside, to do GPU work using F# Interactive scripting on Windows using `dotnet fsi` I ndid the following

* Created DiffSharp.Tests.Gpu that has 

       <DIFFSHARP_TESTGPU>true</DIFFSHARP_TESTGPU>

* Used the following

```
#I "../tests/DiffSharp.Tests.Gpu/bin/Debug/netcoreapp3.0"
#r "TorchSharp.dll"
#r "DiffSharp.Core.dll"
#r "DiffSharp.Backends.Torch.dll"
System.Runtime.InteropServices.NativeLibrary.Load(@"E:\GitHub\dsyme\DiffSharp\tests\DiffSharp.Tests.Gpu\bin\Debug\netcoreapp3.0\runtimes\win-x64\native\torch_cuda.dll")
System.Runtime.InteropServices.NativeLibrary.Load(@"E:\GitHub\dsyme\DiffSharp\tests\DiffSharp.Tests.Gpu\bin\Debug\netcoreapp3.0\runtimes\win-x64\native\nvrtc-builtins64_102.dll")
System.Runtime.InteropServices.NativeLibrary.Load(@"E:\GitHub\dsyme\DiffSharp\tests\DiffSharp.Tests.Gpu\bin\Debug\netcoreapp3.0\runtimes\win-x64\native\caffe2_nvrtc.dll")
System.Runtime.InteropServices.NativeLibrary.Load(@"E:\GitHub\dsyme\DiffSharp\tests\DiffSharp.Tests.Gpu\bin\Debug\netcoreapp3.0\runtimes\win-x64\native\nvrtc64_102_0.dll")
```

gbaydin:
[...]

PR #377, opened 2021-09-10
Improve gpu usage, device api, convolutional and generic vae, unsqueezeAs, unflatten

gbaydin:
These are improvements mainly guided by #204


gbaydin:
Summary of things done in this PR:

- Reorder `(?dtype, ?device, ?backend)` arguments everywhere to `(?device, ?dtype, ?backend)`  because usage such as `.move(device)` is much more common than changing the dtype or the backend.
- Make datasets load data to CPU/RAM by default, regardless the default device and dtype. Loading to the default device/dtype is handled by the dataloader (not datasets). 
- Fix a bug in `bceLoss` that was making it break for tensor shapes other than 2d
- Introduce `unflatten` and `unsqeezeAs` ops
- Introduce a generic `VAE` api that allows things like the following.

```fsharp
let encoder =
    Conv2d(1, 32, 4, 2)
    --> dsharp.relu
    --> Conv2d(32, 64, 4, 2)
    --> dsharp.relu
    --> Conv2d(64, 128, 4, 2)
    --> dsharp.flatten(1)

let decoder =
    dsharp.unflatten(1, [128;1;1])
    --> ConvTranspose2d(128, 64, 4, 2)
    --> dsharp.relu
    --> ConvTranspose2d(64, 32, 4, 3)
    --> dsharp.relu
    --> ConvTranspose2d(32, 1, 4, 2)
    --> dsharp.sigmoid

let model = VAE([1;28;28], 64, encoder, decoder)
```

PR #383, opened 2021-10-10
Upgrade TorchSharp to 0.93.5

gbaydin:
Updating TorchSharp to 0.93.5 that incorporates https://github.com/dotnet/TorchSharp/pull/354 and https://github.com/dotnet/TorchSharp/pull/387 among other things.

@dsyme I removed `LibTorchNugetVersion` from `Directory.Build.props` and everywhere else. We now rely only on the locked in TorchSharp version to also take care of the `libtorch` dependencies. I think this is fine but I wanted to double check with you.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,8 +2,7 @@
 
   <PropertyGroup>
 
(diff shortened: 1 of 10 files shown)


dsyme:
Lovely

PR #388, opened 2021-10-31
Differentiable programs

gbaydin:
This is a minor change in naming and it introduces a `DiffProg<'In, 'Out>` type representing a differentiable program, making it more natural to think about and design general [differentiable programming](https://en.wikipedia.org/wiki/Differentiable_programming) code. 

The `Model` and `Model<'In, 'Out>` types are aliases for `DiffProg<Tensor, Tensor>` and `DiffProg<'In, 'Out>`. The idea with the `Model` types and namespace is that it holds the typical building blocks (or neural network layers, or differentiable programs) conventionally used in deep learning. `Model` corresponds to `torch.nn.Module` in PyTorch.


dsyme:
Should it be "DiffFunc" or similar?  "Program" sounds a little odd, introducing a new concept (Model, Function, Program)

dsyme:
> The Model and Model<'In, 'Out> types are aliases for DiffProg<Tensor, Tensor> and DiffProg<'In, 'Out>. 

Note this might degrade error messages - the alias "DiffProg" may appear.  I'm concerned this will make things more confusing for users

dsyme:
Looking at the code even "DifferentiableObject" might be better.  However it's still hard to see why we'd expose the beginner to the new concept that is identical to Model

gbaydin:
This all comes from some thinking and experimentation with design I've been doing lately. Recently I've reviewed a few papers on differentiable programming language design and I'm working on having this use case expressed very clearly in DiffSharp.

> Should it be "DiffFunc" or similar? "Program" sounds a little odd, introducing a new concept (Model, Function, Program)

I think DiffSharp introduces only concept: a differentiable program. Models are differentiable programs. When/if we talk about functions these should always refer to F# functions, ideally.

I think `DiffFunc` is not the best name because it sounds like a function and I think differentiable programs are different from functions. When I think about a differentiable function I would think about an F# function.

Something like this is a differentiable function. It's an F# function `Tensor -> Tensor` where the output is differentiable with respect to the input.
```fsharp
let f (x:Tensor) = x.sum() + 2.
```

But what differentiable programming in practice refers to is something more than a function. A differentiable program is a collection of functions, parameters, buffers etc. that are tuned under an objective, and after the tuning it's used for a purpose. So it's an object that keeps around this state and can serialize the state etc. 

The following is an example of the type of design that I settled on. This code works in the branch of this PR.

```fsharp
let def p (f:'In->'Out) : DiffProg<'In, 'Out>=
    DiffProg<'In, 'Out>.create [] p [] f

let param (x:Tensor) = Parameter(x)

// Learn a differentiable program given an objective
// DiffProg<'a,'b> -> (DiffProg<'a,'b> -> Tensor) -> DiffProg<'a,'b>
let learn (diffprog:DiffProg<_,_>) loss =
    let mutable l = dsharp.zero()
    for i=0 to 10 do
        diffprog.reverseDiff()
        l <- loss diffprog
        l.reverse()
        // Update diffprog.parameters
    diffprog

// A linear model as a differentiable program
// DiffProg<Tensor,Tensor>
let diffprog =
    let w = param (dsharp.randn([10; 1]))
    def [w] 
        (fun (x:Tensor)  -> x.matmul(w.value))

// let deflinear inSize outSize =
//     let W = param (dsharp.randn([inSize; outSize]))
//     def [W] 
//         (fun (x:Tensor) -> x.matmul(W.value))

// Data
let x = dsharp.randn([1024; 10])
let y = dsharp.randn([1024; 1])

// Objective
// DiffProg<Tensor,Tensor> -> Tensor
let loss (diffprog:DiffProg<Tensor, Tensor>) = dsharp.mseLoss(diffprog.run x, y)

// Learned diferentiable program
// DiffProg<Tensor,Tensor>
let diffproglearned = learn diffprog loss

// Function that runs the differentiable program with new data
// Tensor -> Tensor
diffproglearned.run 
```

gbaydin:
I agree that having `Model` as an alias of something else can confuse the user. But I was thinking it could even be beneficial to make it obvious that "deep learning models are just differentiable programs and differentiable programs are much more general than neural network layers".

One solution could be to not have the type alias at all and just take the step of calling everything `DiffProg` and not using `Model` at all. 

Of course there is the other direction, calling everything `Model` and just forgetting about this differentiable programming concept in the naming. This isn't very good because the model/module name is completely associated with the old typical deep learning concept of NN layers and it really constrains the way people (including me) would think about this capability in the library.

dsyme:
I'd encourage us to talk this through before making this change, maybe on Monday?

I get why you're trying to generalise, but you're also taking something simple and familiar (Models) and effectively making all users learn an abstract concept, but one which 90% of them probably won't care about.  This may turn them off.  It happens all the time in library design (e.g. making people learn "Monad" or "Functor" when all they want to do is print something or transform some data). Most users don't care about this capability. They care about the performance results the capability gives them, or care about using a specific routine someone else has built for them to use this.

The key here is to find a way for advanced users to access the more general functionality without taxing the natural and simple case.  For example an `IDiffObject` interface on Model feels more plausible (advanced users may have to do some downcasting, but that's getting the tax right - a cognitive tax on advanced users, not simple users).

src/DiffSharp.Core/Model.fs
```diff
@@ -3,528 +3,22 @@
...
-
-    /// <summary>TBD</summary>
-    static member load(fileName):Model<'In, 'Out> = ModelBase.load(fileName) :?> Model<'In, 'Out>
-
-    /// <summary>TBD</summary>
-    member m.clone():Model<'In, 'Out> = (m :> ModelBase).clone() :?> Model<'In, 'Out>
 
+/// <summary>Represents a model (a differentiable program mapping a tensor to a tensor).</summary>
+type Model = DiffProg<Tensor, Tensor>
+type Model<'In, 'Out> = DiffProg<'In, 'Out>
```
dsyme, line 17: Why not get rid of `Model<'In, 'Out>`?

gbaydin:
@dsyme and I discussed the ideas explored in this PR and decided to keep the `Model` and `.forward` API. More advanced use cases can introduce the alias `type DiffProg<'In, 'Out> = Model<'In, 'Out>` to emphasize a differentiable programming paradigm when needed. 

In the process of exploring this design, this PR improved some design aspects of model code, especially the code concerning ancestors, descendants, etc. to match graph theory definitions. Another improvement is in the flattening/unflattening of reverse mode tensors in parameter dictionaries. 

Merging this now.

PR #391, opened 2021-11-02
Fix mistake in extensions tests

dsyme:
Hopefully this stays green :)


dsyme:
I'll merge this

gbaydin:
Wow this is actually pretty important. Good catch!

PR #395, opened 2021-12-19
Update to dotnet 6 and F# 6

gbaydin:
This updates dotnet to 6.0. It also adds some useful notes to Directory.Build.props regarding package versioning.

This change broke tar.gz extraction code (used for CIFAR datasets) due to a breaking change in stream behavior in dotnet 6, documented here: https://docs.microsoft.com/en-us/dotnet/core/compatibility/core-libraries/6.0/partial-byte-reads-in-streams

The code is fixed to work correctly on both dotnet 5 and 6.


gbaydin:
Delaying this until fsdocs works with dotnet 6. Merging #412 to move on in the meantime.

gbaydin:
Currently this seems to be blocked due to https://github.com/fsprojects/FSharp.Formatting/pull/719 and https://github.com/ionide/proj-info/issues/122 

I hope these can be solved soon. I'm really excited to move to F# 6 which has some desirable features for diffsharp.

dsyme:
fsdocs-tool 15.0.0 has the update to .NET 6

gbaydin:
Awesome news!

gbaydin:
This is looking good and fsdocs issues seem to be solved. I'm also very happy about the `expr[idx]` notation now allowed. Merging now.

Issue #403, opened 2022-02-08
Strange error regarding "HIP" backend


dsyme:
I just got this strange error running in updated examples/vae.fsx

Code:
```
#I "../tests/DiffSharp.Tests/bin/Debug/net5.0"
#r "DiffSharp.Core.dll"
#r "DiffSharp.Data.dll"
#r "DiffSharp.Backends.Torch.dll"
#r "nuget: TorchSharp-cpu, 0.96.0"
...
<rest of vae.fsx>
printfn "Model\n%s" (model.summary())
```

```
> 
System.Runtime.InteropServices.ExternalException (0x80004005): Could not run 'aten::empty.memory_format' with arguments from the 'HIP' backend. This could be because the operator doesn't exist for this backend, or was omitted during the selective/custom build process (if using custom build). If you are a Facebook employee using PyTorch on mobile, please visit https://fburl.com/ptmfixes for possible resolutions. 'aten::empty.memory_format' is only available for these backends: [CPU, Meta, MkldnnCPU, SparseCPU, BackendSelect, Named, ADInplaceOrView, AutogradOther, AutogradCPU, AutogradCUDA, AutogradXLA, UNKNOWN_TENSOR_TYPE_ID, AutogradMLC, AutogradHPU, AutogradNestedTensor, AutogradPrivateUse1, AutogradPrivateUse2, AutogradPrivateUse3, Tracer, Autocast, Batched, VmapMode].

CPU: registered at aten\src\ATen\RegisterCPU.cpp:16286 [kernel]
Meta: registered at aten\src\ATen\RegisterMeta.cpp:9460 [kernel]
MkldnnCPU: registered at aten\src\ATen\RegisterMkldnnCPU.cpp:563 [kernel]
SparseCPU: registered at aten\src\ATen\RegisterSparseCPU.cpp:959 [kernel]
BackendSelect: registered at aten\src\ATen\RegisterBackendSelect.cpp:609 [kernel]
Named: registered at ..\..\aten\src\ATen\core\NamedRegistrations.cpp:7 [backend fallback]
ADInplaceOrView: fallthrough registered at ..\..\aten\src\ATen\core\VariableFallbackKernel.cpp:60 [backend fallback]
AutogradOther: registered at ..\..\torch\csrc\autograd\generated\VariableType_4.cpp:9226 [autograd kernel]
AutogradCPU: registered at ..\..\torch\csrc\autograd\generated\VariableType_4.cpp:9226 [autograd kernel]
[...]

PR #404, opened 2022-02-08
revert update to TorchSharp 0.96.0

dsyme:
The update to TorchSharp 0.96.0 has some kind of error when doing F# Interactive scripting, see #403 

I'll revert the update and work out what's going on step by step

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -2,7 +2,7 @@
 
   <PropertyGroup>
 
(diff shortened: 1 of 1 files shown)


dsyme:
That bug #403 was due to a problem with F# Interactive in Visual Studio, so we can proceed with the move to TorchSharp 0.96 after all

PR #407, opened 2022-02-09
primitives to allow lifting from operations from TorchSharp tensors to DiffSharp tensors

dsyme:
These primitives allow us to lift primal operations implemented in TorchSharp up to DiffSharp when we can assume the user is using the Torch backend 

@pkese you may be intereted in this

The point is to allow the user to program freely with TorchSharp operations to generate primal tensors.

We use the name `primalRawTorch` (like `primalRaw` but getting a `torch.Tensor`) since it discards differentiability.

### DIFF
--- a/Directory.Build.props
+++ b/Directory.Build.props
@@ -33,7 +33,7 @@
   </PropertyGroup>
 
   <PropertyGroup>
-     <Version>1.0.4</Version>
+     <Version>1.0.5</Version>
      <Authors>Atılım Güneş Baydin, Don Syme, Barak A. Pearlmutter, Jeffrey Siskind, and DiffSharp contributors</Authors>
      <Owners>DiffSharp maintainers</Owners>
      <PackageProjectUrl>https://diffsharp.github.io</PackageProjectUrl>
--- a/src/DiffSharp.Backends.Torch/DiffSharp.Torch.fs
+++ b/src/DiffSharp.Backends.Torch/DiffSharp.Torch.fs
@@ -8,28 +8,21 @@ open TorchSharp
 [<AutoOpen>]
 module Extensions =
 
-    type dsharp with
+    type torch.Tensor with
 
         /// <summary>
-        /// Creates a new tensor from the torch tensor.
+        /// Creates a new DiffSharp tensor from the torch tensor.
         /// </summary>
-        /// <param name="tt">The given TorchSharp tensor.</param>
-        static member ofTorchTensor(tt: torch.Tensor) =
+        member tt.toTensor() : Tensor =
             Tensor.ofRawTensor(TorchRawTensor(tt))
-
-        /// <summary>
-        /// Converts the primal of a tensor to a torch tensor.
-        /// </summary>
-        /// <remarks>If the tensor does not use the Torch backend an exception is raised</remarks>
-        static member primalRawTorch(t: Tensor) =
-            match t.primalRaw with
-            | :? TorchRawTensor as trt -> trt.TorchTensor
-            | _ -> failwith $"primalRawTorch: the input is not a DiffSharp.Backends.Torch tensor, its backend is {t.backend}"
 
     type Tensor with
         /// <summary>
         /// Converts the primal of a tensor to a torch tensor.
         /// </summary>
         /// <remarks>If the tensor does not use the Torch backend an exception is raised</remarks>
-        member t.primalRawTorch() = dsharp.primalRawTorch t
+        member t.primalRawTorch() =
+            match t.primalRaw with
+            | :? TorchRawTensor as trt -> trt.TorchTensor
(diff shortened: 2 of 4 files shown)


dsyme:
@gbaydin I'm finding these necessary for porting some of the Pythong code in the work we're doing.

dsyme:
@gbaydin If it's ok I'll merge this to unblock my porting work? The functions are simple enough that I don't think they need tests.  (We can also adjust the names later)

dsyme:
As an example of code that now ports more straightforwardly, consider
```python
def weight_init_conv2d(out_channels, in_channels, kernel_height, kernel_width):
    w = torch.zeros(out_channels, in_channels, kernel_height, kernel_width)
    nn.init.kaiming_uniform_(w, a=math.sqrt(5))
    return w
```
We want to be able to port this torch code that creates tensors from scratch without much thought, e.g.
```fsharp
let weight_init_conv2d(out_channels: int, in_channels: int, kernel_height: int, kernel_width: int) =
    let w = torch.zeros [| int64 out_channels; int64 in_channels; int64 kernel_height; int64 kernel_width |]
    torch.nn.init.kaiming_uniform_(w, a=Math.Sqrt(5)) |> ignore
    dsharp.ofTorchTensor w
```
This similarly applies to torch input/output and torchvision image manipulation code.

gbaydin:
I'm looking at it now, will comment very soon.

gbaydin:
I think `w.toTensor()` on a torch tensor is not good because w is already a tensor for a torch user.

I think the design should be something like the following. It's ok if these currently don't support differentiation and only move the primal data. This can be mentioned in the docs and that should be good enough.

```fsharp
dsharpTensor = dsharp.fromTorch(torchTensor)
torchTensor = dsharpTensor.toTorch()
```

This is in the same spirit with `torch.from_numpy` that is used to take a numpy ndarray in python and make a torch tensor out of it. I think we shouldn't have anything beyond this simple api.

dsyme:
I'm fine with `dsharp.fromTorch`

For the other one I think there is a fair risk of people (including myself) not realising they're losing differentiability, e.g. say TorchSharp has a tensor function `torch.foo`, then they will do

```fsharp
let foo (x: Tensor) = x.toTorch() |> torch.foo |> dsharp.fromTorch
```
but using that in a model will cause some (likely obscure) error when training, or even worse the differentiation calculation will still succeed under the assumption that the tensor is constant.  Is seems wrong

It seems to me `primalAbc` is currently our consistent terminology to indicate we're losing differentiability and so I'd say stick with that.

dsyme:
Is adding `DiffSharp.Torch` namespace ok?

gbaydin:
> Is adding DiffSharp.Torch namespace ok?

I actually don't like it and would be happy to avoid that if possible.

dsyme:
> I actually don't like it and would be happy to avoid that if possible.

So just add to `DiffSharp`?

gbaydin:
> So just add to DiffSharp?

Yes, if you agree. :)

dsyme:
Done. I changed to `toTorch` as I don't mind.  But be aware it's a trap for people (like converting to numpy is - but it's a bit more subtle as they might be used to the idea that Torch tensors support gradients).

gbaydin:
Thank you!

> Done. I changed to `toTorch` as I don't mind.  But be aware it's a trap for people (like converting to numpy is - but it's a bit more subtle as they might be used to the idea that Torch tensors support gradients). 

I see the point and I agree. If we had some possibility to raise warnings instead of exceptions, this would be a good place to raise a warning. I think F# and dotnet in general doesn't have a mechanism for that if I'm not mistaken. We could print a warning to stderr or something.

dsyme:
>  I think F# and dotnet in general doesn't have a mechanism for that if I'm not mistaken. 

There are various things, including CompilerMessageAttribute.  We should extend that to allow "informational" messages, which are much softer than warnings.

gbaydin:
What PyTorch does for this type of things is to use https://docs.python.org/3/library/warnings.html and keep running without an exception. They frequently use this for things like tensor api that will be deprecated in future versions, warning about problems that can be created by double backward calls in autograd, etc.

dsyme:
> What PyTorch does for this type of things is to use https://docs.python.org/3/library/warnings.html and keep running without an exception. They frequently use this for things like tensor api that will be deprecated in future versions, warning about problems that can be created by double backward calls in autograd, etc.

Yes, in F# you can print warnings, or use `CompilerMessageAttribute`, or the .NET `ObsoleteAttribute`.

gbaydin:
> Yes, in F# you can print warnings, or use CompilerMessageAttribute, or the .NET ObsoleteAttribute.

Awesome, I didn't know!

gbaydin:
This all looks good to me

dsyme:
thanks!

PR #411, opened 2022-02-10
include a DiffSharp-cuda bundle

dsyme:
Similar to https://github.com/dotnet/TorchSharp/issues/512

I'm coming across situations where I'd like a DiffSharp-cuda package that references both DiffSharp-cuda-linux and DiffSharp-cuda-windows, in particular when writing cross-plaftform examples and scripts that require CUDA.

We should just add this, even if it adds GB of download to the initial experience if referenced - it is just too useful when writing samples.


dsyme:
It could be that it's impossible for us to pack this nuget on GitHub Actions due to size limitations, (or we'll need to find some creative way to pack it), let's see what happens

Issue #397, opened 2021-12-20
Types for functions activations used in Sequential?


nhirschey:
Is there interest in a pull request adding a few model layers representing activations for use in Sequential models? Many Pytorch examples use sequential, and this could be helpful for beginners trying to translate pytorch models to diffsharp.

I recognize that you may not want this because it increases the maintenance burden and `-->` covers many `Sequential` use cases. Hence the question.

For example, the pytorch.org [quickstart](https://pytorch.org/tutorials/beginner/basics/quickstart_tutorial.html#creating-models) model definition example uses sequential, and I believe this is the correct translation:

```fsharp
type ReLU() =
    inherit Model()
    override _.ToString() = sprintf "ReLU()"
    override m.forward(value) = 
        dsharp.relu(value)

type Flatten(?startDim:int,?endDim:int) =
    inherit Model()
    let startDim = defaultArg startDim 1
    override _.ToString() = sprintf "Flatten()"
    override m.forward(value) =
        dsharp.flatten(value, startDim, ?endDim=endDim) 
        
type NeuralNetwork() =
    inherit Model()
    let linear_relu_stack = Sequential([
        Linear(28*28, 512)
        ReLU()
        Linear(512, 512)
        ReLU()
        Linear(512, 10)
    ])
    do base.addModel([linear_relu_stack])
    member self.flatten = Flatten()
    
    override self.forward(x) =
        let x = self.flatten.forward(x)
        let logits = linear_relu_stack.forward(x)
        logits


let model = NeuralNetwork()
```

gbaydin:
Hi @nhirschey thank you for this suggestion. 

I'm thinking that introducing `Model` wrapper types around functions is not a good direction for DiffSharp or even for PyTorch. The reason they exist in PyTorch is historical, as far as I know: the `Sequential` and PyTorch module wrappers for simple functions are from the old Torch days when things were composed of coarse-grained neural network modules. When PyTorch introduced general-purpose autodiff, I remember PyTorch devs actually advising against using older containers like `Sequential`, `Parallel` etc.

gbaydin:
I looked at a few ways of implementing the same model you shared above.

```fsharp
open DiffSharp.Compose

type NeuralNetwork2() =
    inherit Model()
    let fc1 = Linear(28*28, 512)
    let fc2 = Linear(512, 512)
    let fc3 = Linear(512, 10)
    do base.addModel([fc1; fc2; fc3])

    override _.forward(x) =
[...]

PR #418, opened 2022-02-21
Model wrappers for functions

gbaydin:
Addresses #397


src/DiffSharp.Core/Model.fs
```diff
@@ -491,12 +491,21 @@ type ModelBase() =
 
 
 /// <summary>Represents a model, primarily a collection of named parameters and sub-models and a function governed by them.</summary>
-[<AbstractClass>]
-type Model<'In, 'Out>() =
+// [<AbstractClass>]
+type Model<'In, 'Out>(?f:'In->'Out, ?parameters: seq<Parameter>, ?buffers: seq<Parameter>, ?models: seq<obj>) =
```
dsyme: Should that `obj` be `ModelBase`?

dsyme:
Seems ok

src/DiffSharp.Core/Model.fs
gbaydin: This design was to improve ease of use for `Model.addModel` etc. but I think it does make sense to get rid of this `obj` and use the correct type.

image

gbaydin: Fixed in a subsequent commit

gbaydin:
I think this is ready to go.

PR #419, opened 2022-02-22
add bias/weight properties To Linear, Conv models

dsyme:
These are already present for BatchNorm* so seems to make sense to add them?


gbaydin:
Sounds good, let's go for it

PR #420, opened 2022-03-06
Covariance and Correlation

nhirschey:
This PR adds covariance and correlation to `Tensor.fs`, `DiffSharp.fs`, and `DiffSharp.Compose.fs`. I am suggesting these additions because I would like to use covariance when using DiffSharp for linear algebra and because PyTorch added them last year.

I followed the code in the PyTorch PRs that added them ([torch.cov](https://github.com/pytorch/pytorch/pull/58311) and [torch.corrcoef](https://github.com/pytorch/pytorch/pull/60420/files)). 

For naming, I used `dsharp.covariance` to follow `dsharp.variance`. I assume `dsharp.variance` instead of `dsharp.var` is a deliberate deviation from torch and numpy naming. But in case it is not deliberate, for what it's worth PyTorch, Numpy, R, and Julia all use `cov` and `var`.

I tried to follow your coding conventions, but if you have any requests or comments, just let me know.

**Examples:**

PyTorch example: https://pytorch.org/docs/stable/generated/torch.cov.html
```fsharp
let x = dsharp.tensor([[0; 2]; [1; 1]; [2; 0]]).transpose()
x.covariance()
(* evaluates to
tensor([[1., -1.],
        [-1., 1.]])
*)
x.covariance(correction= int64 0)
(* evaluates to
  tensor([[ 0.6667, -0.6667],
        [-0.6667,  0.6667]])
*)

let fw = dsharp.tensor([1;6;9])
let aw = dsharp.tensor([0.4282;0.0255;0.4144])
x.covariance(fweights=fw,aweights=aw)

(* evaluates to
tensor([[ 0.4169, -0.4169],
        [-0.4169,  0.4169]])
*)
```

Numpy examples https://numpy.org/doc/stable/reference/generated/numpy.cov.html
```fsharp
let x' = dsharp.tensor([[0;2];[1;1;];[2;0]]).transpose()
x'.covariance()

let a = dsharp.tensor([-2.1;-1;4.3])
let b = dsharp.tensor([3.;1.1;0.12])
let X = dsharp.stack([a;b],dim=0)
X.covariance()
a.covariance()

let xx = dsharp.tensor([[-0.2678; -0.0908; -0.3766;  0.2780],
                        [-0.5812;  0.1535;  0.2387;  0.2350]])

xx.corrcoef()
```


dsyme:
So I guess no differentiation tests need to be added because these are derived, not primitive?

nhirschey:
Yes, they are derived, using pre-existing tensor operations. But after your comment I realized that variance and standard deviation (also derived) have derivatives tests, so I guess you'd probably want covariance and correlation derivatives tested too.

I have now pushed commits with derivative tests for `covariance` and `corrcoef`. Covariance has forward and reverse derivative tests; correlation just has a single reverse test because it is a small transformation of covariance.

- Reverse derivative tests are checked against equivalent code in PyTorch. I left comments with the Pytorch code that I used to get "correct" results.
- I am not sure how you typically check forward derivatives. To check them I wrote functions to calculate the forward derivatives for covariance manually. I left a comment in the test file describing how I'm calculating forward derivatives in those functions. I tried to make it easy to read, but I recognize that checking someone else's manual derivative code is not ideal.

gbaydin:
@nhirschey this is an awesome contribution, thank you. I'm looking at it now and it looks mostly ready to merge.

The naming of `.variance` instead of `.var` wasn't based on a very strong opinion. I think the best thing to do is actually to follow the conventions of all the other systems you listed, so we can use the names `.var`, `.cov` etc. I will modify these in this PR if you don't mind.

nhirschey:
Sounds great, I also like the idea of using var, cov, and std.

gbaydin:
About the testing of forward/reverse derivatives: for these things [difftorch](https://github.com/gbaydin/difftorch) is really useful. It's a little library implementing DiffSharp's differentiation API on top of PyTorch tensors. Calling `.jacobianv` (or `.jvp`) and `jacobianTv` (`.vjp`) operations in difftorch correspond to forward and reverse derivatives.

Issue #422, opened 2022-03-17
In rnn, buffer.nelement differences between newly created stateVector and saved stateVector cause `loadState` to fail


nhirschey:
###  Summary
I don't think that saving and loading state is working properly for RNNs. For example in rnn.fsx, if I create languageModel from scratch there are no buffers, but restoring a previously saved stateVector tries to restore the prior model buffers. The `buffer.nelement` differences between the new model (w/o buffers) and the prior state (w/ buffers) causes an exception when trying to restore the old stateVector. 

I found a workaround saving and restoring "state" rather than "stateVector". Is using "state" preferable?

### Problem

If I run `rnn.fsx` as written, the script uses `languageModel.saveState` to save model state after each `validInterval`.
https://github.com/DiffSharp/DiffSharp/blob/4ab0e66c5b02e1c1c6956f2710ec9f1897fce6f0/examples/rnn.fsx#L89

If I close the session and start again it should reload that state into my model here:
https://github.com/DiffSharp/DiffSharp/blob/4ab0e66c5b02e1c1c6956f2710ec9f1897fce6f0/examples/rnn.fsx#L52

However, if I do that I get an error:
```fsharp
> languageModel.loadState(modelFileName)
System.Exception: Expecting tensors.nelement (855105) and ParameterDict.nelement (854081) to be the same
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1439.Invoke(String message) in D:\a\_work\1\s\src\fsharp\FSharp.Core\printf.fs:line 1439
   at DiffSharp.Model.ParameterDict.unflatten(Tensor tensors)
   at DiffSharp.Model.ModelBase.loadState(String fileName)
   at <StartupCode$FSI_0003>.$FSI_0003.main@()
Stopped due to error
```

This error is caused by `buffer.nelement` differences and the fact that set with `stateVector`  uses `ParameterDict.unflatten` and `ParameterDict.unflatten` requires the new and old nelements to be the same.

For example, summary of a freshly created language model with no buffers:
```fsharp
val it: string =
  "---
Model                                              Params
---
RNNCell(65, 512)                                  295,936
RNNCell(512, 512)                                 524,800
Linear(512, 65)                                    33,345
---
Total params                  : 854,081
Trainable params              : 854,081
Non-trainable params (buffers): 0
---
Total params size (MiB)       : 3.26
---
"
```

Summary of fully restored model saved using `languageModel.save` (*not* saveState):
[...]

Issue #421, opened 2022-03-17
In the rnn.fsx example, languageModel.summary() throws an error after the model has been trained.


nhirschey:
If I run the rnn.fsx example and try to run languageModel.summary() after the training loop, I get this exception
```fsharp
> languageModel.summary()
System.Exception: Cannot cat tensors with different differentiation type (TensorC, TensorF, TensorR).
   at Microsoft.FSharp.Core.PrintfModule.PrintFormatToStringThenFail@1439.Invoke(String message) in D:\a\_work\1\s\src\fsharp\FSharp.Core\printf.fs:line 1439
   at DiffSharp.Tensor.cat(IEnumerable`1 tensors, FSharpOption`1 dim)
   at DiffSharp.Model.ParameterDict.flatten()
   at DiffSharp.Model.ModelBase.summary()
   at <StartupCode$FSI_0014>.$FSI_0014.main@()
Stopped due to error
> 
```

This seems to be caused by `languageModel.buffers` being in reverseDiff mode. If I make the buffers nodiff summary works.
```fsharp
languageModel.state.isNoDiff // true
languageModel.parameters.isNoDiff //true
languageModel.buffers.isNoDiff  //false
languageModel.buffers.isReverseDiff  //true

languageModel.buffers.noDiff() // ()
languageModel.buffers.isNoDiff // true
languageModel.summary()
(*
  "---
Model                                              Params
---
RNNCell(65, 512)                                  295,936
RNNCell(512, 512)                                 524,800
Linear(512, 65)                                    33,345
---
Total params                  : 870,465
Trainable params              : 854,081
Non-trainable params (buffers): 16,384
---
Total params size (MiB)       : 3.32
---
"
*)
```

gbaydin:
It should be fixed in #425. Please let me know if the problem persists.

PR #425, opened 2022-03-31
Fix load and save of models with buffers

gbaydin:
This is to introduce tests and then fix the problem with loading and saving the state of models with buffers, as reported in #421 and #422


gbaydin:
It looks like the main problem behind these errors was introduced in https://github.com/DiffSharp/DiffSharp/commit/33c143df843911b2b80a5c5e6ce54e73f6dba3af

PR #428, opened 2022-04-27
make rnn.fsx example device agnostic.

nhirschey:
Some minor changes that will allow rnn.fsx to run if the dsharp.config is changed to GPU.

Thanks for your incredible work this past month to incorporate my change requests into the dev branch! I'm looking forward to using the new version once it's released on the nuget feed.

### DIFF
--- a/examples/rnn.fsx
+++ b/examples/rnn.fsx
@@ -50,13 +50,14 @@ let modelFileName = "rnn_language_model.params"
 if File.Exists(modelFileName) then 
     printfn "Resuming training from existing model params found: %A" modelFileName
     languageModel.state <- dsharp.load(modelFileName)
+    languageModel.move(Device.Default)
 
 let predict (text:string) len =
     let mutable hidden = rnn.newHidden(1)
     let mutable prediction = text
     let mutable last = text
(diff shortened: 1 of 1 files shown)


examples/rnn.fsx
```diff
@@ -50,13 +50,14 @@ let modelFileName = "rnn_language_model.params"
 if File.Exists(modelFileName) then 
     printfn "Resuming training from existing model params found: %A" modelFileName
     languageModel.state <- dsharp.load(modelFileName)
+    languageModel.move(Device.Default)
```
dsyme, line 53: Shouldn't the load happen on the default device in any case?  Is this a problem with `dsharp.load`?

nhirschey: Yes, I think `dsharp.load` loading into the default device makes sense, and that's what seems to be the goal of PR #430.

Pytorch is a little different, it would torch.load into whatever device the tensor was saved from:  "They are first deserialized on the CPU and are then moved to the device they were saved from" ([link](https://pytorch.org/docs/stable/generated/torch.load.html#torch.load)).

PR #430, opened 2022-05-02
Improve deserialization ease of use

gbaydin:


dsyme:
@gbaydin Do you plan to merge this?

